PR #70, opened 2015-02-08
Prevent type error if property is read only

tobad357:
Some React class exports are read only (react-intl as an example)
If they are then don't redefine them and warn instead

### DIFF
--- a/makeExportsHot.js
+++ b/makeExportsHot.js
@@ -19,9 +19,12 @@ function makeExportsHot(m) {
   for (var key in m.exports) {
     if (freshExports.hasOwnProperty(key) &&
         isReactClassish(freshExports[key])) {
-
-      m.exports[key] = m.makeHot(freshExports[key], '__MODULE_EXPORTS_' + key);
-      foundReactClasses = true;
+      if (Object.getOwnPropertyDescriptor(m.exports,key).writable) {
+      	m.exports[key] = m.makeHot(freshExports[key], '__MODULE_EXPORTS_' + key);
+      	foundReactClasses = true;
+      } else {
+      	console.warn("Can't make class " + key + " hot reloadable due to being read-only");
+      }
     }
   }
 


gaearon:
While this is a fair point, you can also exclude `node_modules` from loader configuration. This will prevent RHL from transforming vendor files:

```js
      { test: /\.jsx?$/, loaders: ['react-hot', 'jsx?harmony'], exclude: /node_modules/ },
```

tobad357:
Well that's an even fairer point, didn't know that was possible :)
If you want I could update the code to warn and suggest excluding as a possible fix?

gaearon:
Yep, let's do that.

Two other nitpicks:

* Indent with two spaces, not one.
* There's a missing space in a function call.

Please squash the changes later.

tobad357:
Would this text be ok?

 Can't make class <<CLASNAME>> hot reloadable due to being read-only, tip: vendor files can be excluded by excluding the /node_modules/ directory from the jsx loader configuration

gaearon:
>Can't make class <> hot reloadable due to being read-only, tip: vendor files can be excluded by excluding the /node_modules/ directory from the jsx loader configuration

I'd change it to:

>Can't make class <> hot reloadable due to being read-only. You can exclude files or directories (such as /node_modules`) with "exclude" option in loader configuration.

By the way, what key is failing? It it trying to process `type`? I'm just concerned if this is something ES6-related that may happen to user components as well.

tobad357:
It's failing on updating module.exports.<>
Ie this doesn't work
module.exports[prop] = ReactClass

Where you are redefining the existing prop.
Doing Object.getOwnPropertyDescriptor(m,'exports',key)
returns that both writable and configurable are false so can't enable writing to the prop either.

Looking at https://github.com/esnext/es6-module-transpiler/blob/master/lib/module.js
Shows that they are setting writable to false and as configurable is false per default we end up with a  situation where the prop can never be changed.

One option could be to redefine the whole exports prop as it is writable?

gaearon:
Oh, I see. Let's open another issue for that.

PR #80, opened 2015-02-16
Update Troubleshooting.md

eiktyrner:
Added some info on how to solve an issue where the users system doesn't have enough watchers available.


gaearon:
Appreciated!

Issue #81, opened 2015-02-16
Problems when `module.exports.__proto__ === null`


moretti:
I'm having some problems updating my project to `react-hot-loader` 1.x because of this particular check in [`makeExportsHot.js`](https://github.com/gaearon/react-hot-loader/blob/master/makeExportsHot.js#L20):
```
if (freshExports.hasOwnProperty(key) &&
```
Essentially my project depends on `cheerio` and one of its dependendencies, [`css-select`](https://github.com/fb55/css-select), is [exporting objects with `__proto__ === null`](https://github.com/fb55/css-select/blob/master/lib/general.js#L12). 

I really think that redefining `__proto__` is a bad idea, but perhaps we could make the check more robust against such cases.

What do you think?

Thanks! :wink:

gaearon:
Is the check failing or something? Can you submit a PR to fix it?

gaearon:
Thanks! By the way, while it's good to have this fix in, you can just set `exclude: /node_modules/` in your loader config: https://github.com/gaearon/react-hot-boilerplate/blob/master/webpack.config.js#L24

This way we won't process any your dependencies.

PR #82, opened 2015-02-16
Fix for modules with null [[Prototype]] chain

moretti:
This is a PR for https://github.com/gaearon/react-hot-loader/issues/81

A module created in the following ways:
```
module.exports = Object.create(null)
// or
module.exports = { __proto__: null }
```
won't have `hasOwnProperty()` available on it.

So it’s safer to use `Object.prototype.hasOwnProperty.call(myObj, prop)` instead of `myObj.hasOwnProperty(prop)`.

### DIFF
--- a/makeExportsHot.js
+++ b/makeExportsHot.js
@@ -17,7 +17,7 @@ function makeExportsHot(m) {
   }
 
   for (var key in m.exports) {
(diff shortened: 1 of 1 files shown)


moretti:
Thanks! :+1:

gaearon:
Out in 1.1.6. Thank you!

PR #102, opened 2015-03-17
use escaped unicode for property values when using content

moimikey:
*snort*

renders most of the time, but i've had this happen to me when checking back:

screen shot 2015-03-17 at 3 13 52 pm


gaearon:
wow

Issue #64, opened 2015-02-02
Opt into unmounting for a specific component for some period of time


gaearon:
Sometimes I edit `componentDidMount` or `getInitialState` and I actually want the component to re-mount so these changes take effect. After I'm done with these, I want hot reloading to work as usual.

I couldn't think of a non-obtrusive way to do that yet. If opting into this behavior takes more than a few seconds, I'd rather just hit "Refresh".

Things I'm considering is some “magic” property on component, or a “magic” comment in the file a la `/* cold */` that turns on this behavior.

Any ideas?

syranide:
I usually find it easiest to just "go back out and in again" to accomplish that. An interesting possibility would be to remount components if the keys returned from a call to `getInitialState` differs from the initial mount (but then you'd have to track that).

gaearon:
I was thinking about special casing `getInitialState` but that's too fragile. It may return different keys because store state is different.

I think I'll add some sort of magic comment as an experiment and see whether it helps my workflow. Not a big deal.

KyleAMathews:
I just refresh. I find that rhl gets the code into a weird state sooner or later so have to refresh anyways after enough code edits.

tcoopman:
I also refresh. It would be nice though if you could replay all your actions after an refresh:
Seems like a nice use case for https://github.com/gaearon/flux-hot-loader :)

sokra:
Maybe an optional `stateVersion` property on the component. Everytime it changes state is invalid and component should be remounted...

patrickheeney:
New user to react and hot loader. I have been working on getting this setup for about a day. I incorrectly assumed that state changes should have something happen. I kept debugging with the assumption that this was not standard behavior. I would love a troubleshooting guide or some notes somewhere regarding how react-hot-loader handles state and when we should expect it to refresh.

gaearon:
I think I should split `docs/README` into several files in `docs` and add `Troubleshooting` there. Link to `Troubleshooting` from Get Started on landing, main README and from React Hot Boilerplate README. Makes sense?

kibin:
Any updates on this?

gaearon:
I agree refreshing works fine in these cases and couldn't think of a more obvious way to do that. I'm still open to suggestions tho

PR #103, opened 2015-03-21
Support custom dispose function

arsemyonov:
Hey guys,

This PR is very experimental, suitable only for **discussion** and not a candidate for merge. Related to #64. Proof of concept.

Well, since I'm too lazy guy, I decided to try to implement something that will give me ability to mark some components/modules as `ready to be force reloaded`.

What are these modules? For instance, I could have like following:

```javascript
React.createClass({
  componentDidMount() {
    jQuery(this.refs.container.getDOMNode()).greatlyImplementedChartLibrary();
  },
  ...
})
```

As I said above, I'm lazy to refresh the page and wait, I want to reload such components automatically. So, how I tried to solve it (I use `react-router`)

  1. Wrapped ReactRouter (not required, I did it just for simpler singleton usage). Up to you. I think we can play router's `transitions`. Haven't tried.
  2. Defined custom `dispose` function in webpack.conf.js (not `function`, but `string` actually).
  3. Slightly changed `react-hot-loader`

So,

```javascript
// webpack.conf.js

var disposer = [
  'function() {',
    'var router = require(' + JSON.stringify(require.resolve('app/router.js')) + ');',
    'router.restart();',
  '}'
].join('\n');

module.exports = {
  module: {
    loaders: [{
      test: /.jsx$/,
      loader: 'react-hot',
      query: {
        disposer: disposer
      }
    }}
  }
}
```

```javascript
// router.js
import React from 'react';
import BaseRouter from 'react-router';
import routes from './routes';

class Router {
  constructor(container, handlerGetter) {
    this.router = BaseRouter.create({ routes });
    this.handler = handlerGetter(container);
    this.container = container;
  }

  start() {
    this.router.run(this.handler);
  }

  stop() {
    this.router.stop();
    React.unmountComponentAtNode(this.container) // + this.start() → will blink
  }

  restart() {
    this.stop();
    setTimeout(() => {
      this.start();
    }, 1);
  }
}

export default new Router(document.getElementById('app'), (container) => {
  return (Handler) => {
    React.render(<Handler />, container);
  }
});
```

```javascript
// SomeComponent.jsx
var SomeComponent = React.createClass({
  componentDidMount() {
    jQuery(this.refs.container()).greatlyImplementedChartLibrary();
  },
  ...
});

if(module.hot) {
  module.forceReload = true;
}

module.exports = SomeComponent;
```

Of course, it will be much simpler If you don't use `react-router`.

I don't think this could be merged ever, but it can be as an idea/inspiration. Anyway, thanks for your time and sorry for this shit.

Cheers,
Art

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,9 +3,14 @@
 var path = require('path'),
     SourceNode = require('source-map').SourceNode,
     SourceMapConsumer = require('source-map').SourceMapConsumer,
-    makeIdentitySourceMap = require('./makeIdentitySourceMap');
+    makeIdentitySourceMap = require('./makeIdentitySourceMap'),
+    loaderUtils = require("loader-utils");
+
 
 module.exports = function (source, map) {
+  var query = loaderUtils.parseQuery(this.query);
+  var disposer = query.disposer || (function() {}).toString();
+
   if (this.cacheable) {
     this.cacheable();
   }
@@ -47,6 +52,7 @@ module.exports = function (source, map) {
       '(function () {',
         'module.hot.dispose(function (data) {',
           'data.makeHot = module.makeHot;',
+          'module.forceReload && (' + disposer + ')()',
         '});',
 
         'if (module.exports && module.makeHot) {',
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Tweak React components in real time.",
   "main": "index.js",
   "dependencies": {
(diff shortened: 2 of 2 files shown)


gaearon:
We're in the process of simplifying RHL a lot, so such experimentation will be easier to do in userland.
Closing for now, thanks for submitting a PR!

PR #105, opened 2015-03-22
Prevents makeHot() twice on a React 0.12 component

RickWong:
Take this example:

````
var C = React.createClass(...);
module.exports = C;
````

Currently React Hot Loader will call `makeExportsHot(module.exports, React)` where not only `makeHot(C, '__MODULE_EXPORTS')` is performed, but also by property-traversal `makeHot(C.type, '__MODULE_EXPORTS_type')`. The issue with this is upon modifying and hot-loading C, the component is actually updated twice and React lifecycle methods like `render` are also executed twice.

This patch fixes that redundant behavior.


gaearon:
This is good, thanks! I'll get it released on Monday.

PR #109, opened 2015-03-25
Add yarsk to starter kits

bradleyboy:
Hi - I have a simple starter kit known as **yarsk** (Yet Another React Starter Kit) that has hot-loading enabled out of the box.


gaearon:
Awesome!

PR #112, opened 2015-03-30
use include instead of exclude

tcoopman:
See https://github.com/gaearon/react-hot-loader/commit/5d5b356bff510dd6c4e1a3838134a087206051c5


gaearon:
Appreciated, thanks.

Issue #114, opened 2015-04-07
Same-origin policy using hot loader inside an iframe.


gdborton:
I'm getting the below error when running webpack-dev-server and react-hot-loader inside of an iframe of a different protocol/port/domain.

My use case is parent: localhost:4100 and child: localhost: 3000

```
Uncaught SecurityError: Failed to read the 'frame' property from 'Window':
Blocked a frame with origin "https://localhost:3000" from accessing a frame with
origin "https://localhost:4100". Protocols, domains, and ports must match.
```

Updating [this line](https://github.com/gaearon/react-hot-loader/blob/master/makeExportsHot.js#L32) to the below, makes it work for me.
`if (key === 'frameElement' || !isReactClassish(freshExports[key], React)) {`

gaearon:
Can `try/catch` help here? We wouldn't want to special-case any specific properties that might throw in some rare cases.

gdborton:
That should work as well.  Did you want to do anything specific with the exception?

gaearon:
I think we may silently ignore it, or maybe warn unless it's too noisy.

gaearon:
Should be fixed in 1.2.9.

PR #115, opened 2015-04-08
Add try/catch to makeExportsHot.

gdborton:
Addresses the issue I opened here - #114

### DIFF
--- a/makeExportsHot.js
+++ b/makeExportsHot.js
@@ -29,9 +29,14 @@ function makeExportsHot(m, React) {
       continue;
     }
 
-    if (!isReactClassish(freshExports[key], React)) {
+    try {
+      if (!isReactClassish(freshExports[key], React)) {
+        continue;
+      }
+    } catch (e) {
       continue;
     }
+
 
     if (Object.getOwnPropertyDescriptor(m.exports, key).writable) {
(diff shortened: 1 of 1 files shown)


makeExportsHot.js
```diff
@@ -29,10 +29,15 @@ function makeExportsHot(m, React) {
...
-    if (!isReactClassish(freshExports[key], React)) {
+    try {
+      if (!isReactClassish(freshExports[key], React)) {
+        continue;
+      }
+    } catch (e) {
       continue;
     }
 
+
```
gaearon: Nit: extra newline

makeExportsHot.js
```diff
@@ -29,10 +29,15 @@ function makeExportsHot(m, React) {
       continue;
     }
 
-    if (!isReactClassish(freshExports[key], React)) {
+    try {
```
gaearon: I think it would be nicer to put `try`/`catch` around  something like `var maybeReactClass = freshExports[key]` to make it clear where we might expect the error.

gaearon:
Thanks for the effort. Superseded by https://github.com/gaearon/react-hot-loader/commit/64dd55a0cca4b4a117b087053601489a2f9ddaf0.

PR #116, opened 2015-04-10
Pass React to isReactElementish

matthewwithanm:
I was tracking down a weird problem and in the course of it, got an error saying that React was undefined [here][1]. It turned out that it was because `isReactElementish` doesn't pass `React` to `isReactClassish` when it calls it [here][2].

My weird error ended up being unrelated to this and in normal operation, I never seem to get to this branch, but it seems like `React` isn't an optional argument to `isReactClassish` and therefore needs to be given to the stuff that calls it (like `isReactElementish` too) so I figured I'd open this PR (:

[1]: https://github.com/gaearon/react-hot-loader/blob/ea86b4684b7846d82fe6c11e46c93f2bc55d2b49/isReactClassish.js#L11
[2]: https://github.com/gaearon/react-hot-loader/blob/ea86b4684b7846d82fe6c11e46c93f2bc55d2b49/isReactElementish.js#L9

### DIFF
--- a/isReactElementish.js
+++ b/isReactElementish.js
@@ -1,12 +1,12 @@
 var isReactClassish = require('./isReactClassish');
 
-function isReactElementish(obj) {
+function isReactElementish(obj, React) {
   if (!obj) {
     return false;
   }
 
   return Object.prototype.toString.call(obj.props) === '[object Object]' &&
-         isReactClassish(obj.type);
+         isReactClassish(obj.type, React);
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks, good catch.

gaearon:
Landed in 1.2.5

matthewwithanm:
:thumbsup: No problem!

PR #136, opened 2015-05-22
Remove NoErrorsPlugin

gaearon:
This is something silly I've been doing and recommending for a long time. Sorry! There's no need for `NoErrorsPlugin` if you use `hot/only-dev-server`. Hot updates won't blow your app away if you make a syntax error, but at least you'll see the error in the DevTools console! And the subsequent hot updates will just work when you fix that error.

### DIFF
--- a/getstarted.md
+++ b/getstarted.md
@@ -37,7 +37,7 @@ In your <a href="https://github.com/gaearon/react-hot-boilerplate/blob/master/we
 {% highlight js %}
 entry: [
   'webpack-dev-server/client?http://0.0.0.0:3000', // WebpackDevServer host and port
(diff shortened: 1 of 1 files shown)


gaearon:
Oh wait, I acted too fast :-/

This seems to break HMR if I `import` a non-existent file.
I'm going to ask around in Webpack repo because I really want errors in console *and* I want imports to not break HMR.

gaearon:
So, `NoErrorsPlugin` is not needed after all!

>This seems to break HMR if I import a non-existent file.

was fixed by https://github.com/gaearon/react-hot-loader/pull/187.

PR #138, opened 2015-05-24
Replace "see" with "watch" in github page

cesarandreu:
Currently it says "[See the video](https://www.dropbox.com/s/cv58f0jlslopznz/Screenshot%202015-05-24%2013.18.05.png?dl=0)", but I think it makes more sense for it to be "Watch the video". 

[This link](https://www.englishclub.com/vocabulary/cw-see-look-watch.htm) seems like a good reference for when to use "see", "look", or "watch".

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -1,7 +1,7 @@
 <nav class="nav">
   <ul class="nav-list nav-list-normal">
     <li class="nav-list-item">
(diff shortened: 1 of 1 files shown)


gaearon:
Does it fit on every media query as you resize the window down?

gaearon:
Seems good, thanks!

PR #152, opened 2015-07-01
add troubleshooting osx fsevent bug

dunckr:
I actually had this obscure OSX bug when trying to get react-hot-loader working. Hopefully will help others get up and running quicker than me 😄


gaearon:
:+1:

PR #155, opened 2015-07-07
Update Troubleshooting.md

unknownexception:
This document has one searchable issue but no solution provided. Fixed that.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -60,6 +60,18 @@ Most likely you used `npm link` to use a development version of a package in a d
 ### Page Throws an Error
 
 #### Uncaught TypeError: Cannot read property 'NODE_ENV' of undefined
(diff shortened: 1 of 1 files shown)


gaearon:
Are you sure this is related to React Hot Loader? I don't remember us using `process.env`.

unknownexception:
It's more related to `webpack`, you're right. For some reason, the exception was already in `Troubleshooting.md`:
`Uncaught TypeError: Cannot read property 'NODE_ENV' of undefined`
And it's on the first position at Google, so it's just confusing. Now I get it, this is because https://github.com/mjohnston/react-native-webpack-server/issues/34, and the right solution is addingg to whitelist only ES6 node_modules in `loaders`, not including `react`.

gaearon:
Yes, it's on the page because it corresponds to the next section.
Both titles are for the same section that tells to exclude `node_modules`.

PR #179, opened 2015-08-18
add react as peerDependencies

sorrycc:
so react can be installed automatically.


gaearon:
See why we removed it: https://github.com/gaearon/react-hot-loader/pull/146. :-)

PR #181, opened 2015-08-22
Check if module exists before using it.

trbngr:
I realize this is an edge case, but I ran into a situation recently. 

I needed some static, non-commonjs, JS to be put in the output directory using file-loader, and react-hot-loader kept injecting itself, and therefore, causing runtime errors.

The solution was simple. Instead of adding config to the loader to ignore stuff, I simply check for module before using it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ module.exports = function (source, map) {
 
   prependText = [
     '/* REACT HOT LOADER */',
-    'if (module.hot) {',
+    'if ("object"==typeof module && module.hot) {',
       '(function () {',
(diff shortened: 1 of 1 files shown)


trbngr:
...and I just realized that I could exclude the static files directory ;)

gaearon:
>...and I just realized that I could exclude the static files directory ;)

This :-).

We're assuming we're inside Webpack so `module` is available, otherwise we'd also need to check for `require`, etc.

gaearon:
Actually it's always better to `include` just the source directory instead of `exclude`-ing something.

PR #187, opened 2015-08-28
Recover from errors on module level

gaearon:
Currently, if the module definition throws, even if you fix the problem, hot reloading will no longer work. This is because an error, if any, occurs before `module.hot.accept()` call, and it never executes, so module isn't marked as safe to reload.

This change should fix the problem. We're now wrapping module code in `try`/`finally` (note: no `catch`, we just want to execute `module.hot.accept()` inside `finally` whether the code failed or not).


index.js
```diff
@@ -37,36 +37,37 @@ module.exports = function (source, map) {
...
+          'if (module.exports && module.makeHot) {',
+            'var makeExportsHot = require(' + JSON.stringify(require.resolve('./makeExportsHot')) + ');',
+            'if (makeExportsHot(module, require("react"))) {',
+              'foundReactClasses = true;',
+            '}',
+            'var shouldAcceptModule = ' + JSON.stringify(acceptUpdates) + ' && foundReactClasses;',
+            'if (shouldAcceptModule) {',
+              'module.hot.accept(function (err) {',
+                'if (err) {',
+                  'console.error("Cannot not apply hot update to " + ' + JSON.stringify(filename) + ' + ": " + err.message);',
```
doctyper: > Cannot not

Double negative?

gaearon: Lol. It has been that way forever and I never noticed it..

gaearon:
Note: this means you can finally remove `NoErrorsPlugin` and enjoy Babel errors in your console, as well as hot reload code after fixing those errors.

PR #197, opened 2015-09-17
Add deprecation notice

gaearon:
React Hot Loader is going to be phased away in favor of the [React Transform](https://github.com/gaearon/react-transform-boilerplate) family of projects.
We will soon have a migration guide and a detailed overview, but in the meantime, feel free to check them out and play with them!


Andersos:
So why is react-hot-loader deprecated in favor of React Transform?

PR #201, opened 2015-10-04
Update Troubleshooting.md

arturi:
I’ve struggled with this issue for a few hours trying to implement https://github.com/gaearon/react-transform-boilerplate in my existing Express app. A note about ```.babelrc``` might help future padavans.


gaearon:
Thanks, but React Hot Loader is unrelated to React Transform. They both use HMR API so warnings are similar, but this isn't really the right place for this tip.

arturi:
All right, maybe place it anywhere else you think is appropriate, just someone will be able to google, like I did that troubleshoot page.

PR #206, opened 2015-10-19
Add React Typescript Boilerplate.

keokilee:
Add another react-typescript-boilerplate to examples. This actually uses React Transform HMR, so I'm not actually sure if this is the right place or not. Especially since this repo is going to be deprecated.


gaearon:
Sorry for the turbulence. We’ll unite both projects eventually but for now this isn’t relevant to this repo.

PR #220, opened 2015-12-29
Added additional instructions for modules and HMR

eblanchette:


eblanchette:
Added additional instructions for HMR and webpack modules

PR #236, opened 2016-03-15
Test also properties of module.exports.default

alfonsogarciacaro:
I know this is not frequent but it may happen that React components are exposed as properties of a default export (as it occurs, for example, when compiling F# to JS with [Fable](https://github.com/fsprojects/Fable)). This PR just adds an additional test on members of `m.exports.default` when no React classes have been found for `m.exports`.

Thanks a lot for this wonderful tool, Dan!


gaearon:
Seems like you got it solved there instead!

alfonsogarciacaro:
Yeah, sorry I forgot to close the PR :/ Thanks for that!

PR #240, opened 2016-04-17
React Hot Loader 3

gaearon:
Work in progress.
(We are skipping 2 because the approach is way too different from 2 alpha which went nowhere).

Some info here: https://github.com/gaearon/react-hot-boilerplate/pull/61


nfcampos:
@gaearon there's a bug with the babel plugin, it thinks variables babel creates to transpile arguments spread should be tagged, see this gist https://gist.github.com/nfcampos/216edf34ae47dfff809df41c58b58c8f

anaibol:
What's the diference between React Hot Loader and React transform / HMR? I thought that React Hot Loader was dead in favor of React Transform.

gaearon:
@anaibol 

Thank you for asking! I go over this difference in detail here: https://medium.com/@dan_abramov/hot-reloading-in-react-1140438583bf. The result (React Hot Loader 3) is what I came up with, having learned from both approaches and their weaknesses.

gaearon:
@nfcampos Thanks for reporting, should be fixed in `-alpha.9`.

src/AppContainer.dev.js
```diff
@@ -0,0 +1,109 @@
...
+      error: null
+    });
+  }
+
+  componentDidUpdate(prevProps) {
+    // Hot reload has finished.
+    // Force-update the whole tree, including
+    // components that refuse to update.
+    if (prevProps !== this.props) {
+      deepForceUpdate(this);
```
nfcampos: @gaearon calling `deepForceUpdate` in `componentWillUpdate` makes the children component re-render twice every time it changes, one due to the regular react lifecycle, another due to the `deepForceUpdate`. If we moved `deepForceUpdate` to `componentWillReceiveProps` it would re-render only once. Do you see any downside?

gaearon: If it works correctly, sure.

ChrisCates:
@gaearon Is this ready for testing?

gaearon:
I think it is ready but it’s hard to say until it gets more usage.

glenjamin:
@gaearon do you have any notes which explain why the React Router hack isnt needed anymore? I can't see any difference from when the original issue was hit?

gaearon:
@glenjamin This is the relevant commit: https://github.com/gaearon/react-hot-loader/pull/272/commits/e41e11701754c5026b36d286495bd0ff1bc27777. Relevant PR: https://github.com/gaearon/react-hot-loader/pull/272. Rather than assign `_source`, we now call `.register()` from the modules, so RHL doesn’t care whether or not the new types get `createElement`ed. The new code updates proxies during module definition (`register()` call) rather than during first use (`createElement()`).

glenjamin:
I was about to write something about how I didn't understand how that differed from the pre-v3 approach, but I think I have it now:

## Pre v3
* Attempt to detect type of thing at definition time
* Replace definition in-line with proxy
* createElement is normal

## v3 alpha
* Tag every export of every module
* Leave the original definition alone
* create/update proxies by tag when `createElement` is called

## v3
* Tag every export of every module
* Leave the original definition alone
* Create/update proxies in a global map based on tag at definition time
* Resolve to the current proxy when `createElement` is called


That look about right?

gaearon:
Yes!

gaearon:
Also, in v3, “Tag every export of every module” depends on whether you use Webpack loader or Babel plugin. Webpack loader tags everything in `module.exports` but Babel plugin tags any top-level variable, so it works with unexported classes or HOC call results.

src/AppContainer.prod.js
```diff
@@ -0,0 +1,18 @@
...
+
+'use strict';
+
+const React = require('react');
+const { Component } = React;
+
+class AppContainer extends Component {
+  render() {
+    if (this.props.component) {
+      return <this.props.component {...this.props.prop} />;
```
leidegre: Should be `this.props.props` and not `this.props.prop`, no?

glenjamin: Is it worth dropping the `component` prop and only having `children` before this goes into a final release perhaps?

nfcampos: @glenjamin yeah I think that's the idea

leidegre: Curious. Why does both ways exist?

nfcampos: because the first alphas had the `component` + `prop` way and later it was changed

leidegre: I'll switch over to use children as a primary workaround for the bug and I think if the component/props way gets dropped that's just better. Assuming it has no other purpose.

nfcampos: Yep do switch to `children`. The other way is there just for backward compatibility now

glenjamin:
Is this warning still relevant? https://github.com/gaearon/react-hot-loader/pull/240/files#diff-8be7b0cf5cd0de13e69f70bcf3a11143R90

What stopped `createElement` from getting the right component in this case originally?

gaearon:
>What stopped createElement from getting the right component in this case originally?

Something like this:

```js
class App {
}

ReactDOM.render(<App />) // Whoops, early element usage!

(function () {
  __REACT_HOT_LOADER__.register(App) // Executes too late
})
```

Since the initial render was done with vanilla element, registering a proxy now would cause it to unmount on the next render. This can be unexpected, so we just bail out.

glenjamin:
> Since the initial render was done with vanilla element, registering a proxy now would cause it to unmount on the next render, as we would start returning a proxy next time.

Ah, that makes sense.

Would it be possible to place the register call immediately after the definition? That way there wouldn't be a dead zone between definition and registration.

gaearon:
>Would it be possible to place the register call immediately after the definition? That way there wouldn't be a dead zone between definition and registration.

It would be possible for Babel plugin but not for Webpack. To be honest I think it’s a fair compromise because it makes the plugin *really* non-invasive. It just appends some stuff at the bottom, so it’s safe to say it’s not fragile. And it’s easy to refactor this to the desired pattern anyway.

davidpfahler:
@gaearon As facebook/react#6020 seems to not have landed yet, I assume hot reloading does not yet support error reporting in RHL 3, right? Also, do you intend to embed `redbox-react` in the RHL 3 error reporting (some of your comments seem to suggest that)? I tried to to move the [RHL example of `redbox-react` to 3.0.0-beta.2](https://github.com/KeywordBrain/redbox-react/pull/59), but the error is not being caught and I don't know how to hand it over to redbox. Any suggestions?

ntucker:
I get the redbox error when I first load a page with an error, but whenever it is fine, and then I save an error, I get in the console:

[HMR] Cannot apply update. Need to do a full reload!

And all hotreloading no longer works. Here's my config modifications for dev mode

config.entry.App = [
  'react-hot-loader/patch',
  'webpack-dev-server/client?http://localhost:3000/',
  'webpack/hot/only-dev-server',
  ...config.entry.App,
]
config.plugins = [
  new webpack.HotModuleReplacementPlugin(),
  new webpack.NoErrorsPlugin(),
  ...config.plugins,
]
config.devServer = {
  hot: true,
  port: 3000,
}

davidpfahler:
@ntucker In order to get error catching on save with RHL 3 you'd need to build [this react PR](https://github.com/facebook/react/pull/6020) yourself and use it instead of latest react. We hope these changes land soon in latest react.

ntucker:
I'm getting 

> [react-router] You cannot change <Router routes>; it will be ignored

How exactly do I configure react-router with this?

sarovin:
@ntucker i'm trying. See this --> http://stackoverflow.com/questions/37466598/react-hot-loader-3-with-webpack-dev

ericclemmons:
Dunno if anyone else has this issue, but...

When using Webpack with `devtool: "cheap-eval-source-map"` and `react-hot-loader/babel`, I get the following error (even in an empty file!):

```
ERROR in ./src/client.js
Module build failed: TypeError: /Users/Eric/Projects/HigherEducation/platform/src/widgets.js: Cannot assign to read only property 'mappings' of {"version":3,"sources":[],"names":[],"mappings":"","sourcesContent":[]}
```

It has the same problem with `inline-sourcemap` as well, so it must be issue with using sourcemap?

gaearon:
>I'm getting
>[react-router] You cannot change ; it will be ignored
> How exactly do I configure react-router with this?

This warning message is absolutely harmless; you can ignore it. Unfortunately there is no way to opt out of it at the moment (you can contribute to fix it in https://github.com/reactjs/react-router/issues/2182).

qrohlf:
When I switch my pure functional class to an es6 class, hot reloading no longer works and everything then requires a full reload.

working, hot
```
import React from 'react'
import staticProps from './props'
import MultiRouteEmbed from './components/MultiRouteEmbed'

export default function App (props) {
  return (
    <div>
      <MultiRouteEmbed {...staticProps} />
    </div>
  )
}
```

no longer hot
```
import React from 'react'
import staticProps from './props'
import MultiRouteEmbed from './components/MultiRouteEmbed'

export default class App extends React.Component {
  render () {
    return (
      <div>
        <MultiRouteEmbed {...staticProps} />
      </div>
    )
  }
}
```

is this expected behavior?

ChrisCates:
Things looking great so far @gaearon keep up the good work buddy! 
I'm blessed to have people like you make my development career a better experience.

MoOx:
Is this normal that `componentWillMount` is called for components behing "hot reloaded"?
If not, I am having stateful (class) components rendered from stateless (function) components, can this be the cause of this behavior?

For example I currently have some data fetched in this part of the lifecycle and so I am getting new fetch() call each time I update the render, which is not optimal in my case (but I can understand it can be convenient in some other cases tho).

leidegre:
@MoOx it's work in progress. I think I see the same thing. I don't think it's supposed to be like that but the intricacies of why that is eludes me. You should however see a corresponding `componentWillUnmount` and if you're component is built proper it should be able to deal with it even though it loses internal state. It's unfortunate but the current state of things.

nathanmarks:
@MoOx @leidegre 

They shouldn't be mounting/unmounting unless that's a result of one of your own `componentWillUpdate` functions. Mine are just running the update hooks with the new hot loader.

leidegre:
@nathanmarks What exactly is the *new* hot loader? What packages which versions? I'm using `react-hot-loader@3.0.0-beta.2`.

leidegre:
@nathanmarks are you using `react-router` but I think the reason I'm seeing this is because I'm using a hack to force React.createElement to happen.

nathanmarks:
@leidegre Yeah, `3.0.0-beta.2` is the new hot loader. The `AppContainer` uses `forceUpdate` -- unless I've missed something, doesn't that only trigger an update? I'm not sure everything remounting is supposed to happen. I'm using it with `react-router` as well btw.

gaearon:
>Is this normal that componentWillMount is called for components behing "hot reloaded"?

It’s not. Please provide an example reproducing this.

>I am having stateful (class) components rendered from stateless (function) components, can this be the cause of this behavior?

No, this shouldn’t remount inner components. But I can’t help without a reproducing case.

>It's unfortunate but the current state of things.

And it will stay the current state unless you can share the example reproducing this or willing to look into why this happens 😉 .

gaearon:
>are you using react-router because I think this is the reason I'm seeing this. I'm using a hack to force React.createElement to happen.

What hack? There was a mention of a hack before (with alphas) but it is not necessary with betas.

leidegre:
@gaearon I'll slap together a repo and see if I cannot reproduce this or if it's just my perception of the situation that is off.

The hack I'm refering to is this react-router hack https://github.com/gaearon/react-hot-boilerplate/pull/61#issuecomment-211504531

leidegre:
@gaearon Here you go https://github.com/leidegre/react-hot-loader-beta.2-test

This is not just the react-hot-loader it's a lot of stuff, but I get a unmount/mount to happen if I just change client/chrome/welcome.jsx and save the changes. You can see this if the console window is open when the update is pushed to the browser.

I've fooled around with toggling things off. I have so far just assumed that it is the react-router hack that's the culprit but I removed that and it's still happening.

Dharmoslap:
Will be this working with TypeScript as well? Our team is using it instead of babel.

nathanmarks:
@leidegre I don't think you can use a functional component for the hot root. @gaearon can confirm though

Kovensky:
@nathanmarks functional root components work fine (as long as you _never ever_ hot accept them in the same module as they are defined)

The problem @leidegre is having (*assuming webpack2, but hopefully regular babel/ts transforms are the same) is not using ES6 modules correctly. When you use "export default", you create a static (i.e. not live!) binding, and the commonjs re-import is unnecessary when the modules are done correctly.

```js
// root.jsx
// note: NO default!
export function Root (props) { ... }
```

```js
// index.ts
import { Root } from './root'

render(<AppContainer><Root store={store} /></AppContainer>)

if (module.hot) {
  // no require, Root is a live binding, so you can just use it directly
  module.hot.accept('./root', () =>
    render(<AppContainer><Root store={store} /></AppContainer>))
}
```

If you really want to be able to just `import Root from './root'` while keeping it a live binding, change `root.jsx` to:

```js
export { Root as default }
function Root (props) { ... }
```

nathanmarks:
@Kovensky thanks for the clarification!

wmertens:
@Dharmoslap did you try using the webpack plugin? That should work…

thecentric:
I'm having an issue with this branch when trying to run on Ubuntu 14.04 (working fine for a colleague on OSX).

    ERROR in ./src/index.js
    Module not found: Error: Cannot resolve module 'react-hot-loader' in /home/demelza/Desktop/react-hot-boilerplate/src
     @ ./src/index.js 3:22-49

Any ideas?  I tried a couple of versions of node.

MoOx:
@thecentric you probably don't have the dependency installed. Try "npm install" if the deps is already specified in your package.json, or "npm install --save-dev react-hot-loader@^3.0.0-beta.2" if not.

thecentric:
@MoOx thanks for your reply, unfortunately that's not working; I can see the package in `node_modules` so it's definitely installed (I tried the second command you gave anyway, but no luck).

dazld:
I realise the majority of users will be using webpack, but is there somewhere I can see an example of how this would work with gulp & browserify?

tsnieman:
@thecentric Look closely at your error message:

> ERROR in ./src/index.js
> Module not found: Error: Cannot resolve module 'react-hot-loader' **in /home/demelza/Desktop/react-hot-boilerplate/src**
>  @ ./src/index.js 3:22-49

It's only looking for the module in the `src` directory. I think you need to add `node_modules` to your module resolution paths.

vinhlh:
When building bundles, there are too much code like this
```
constants/actionTypes.js"), __REACT_HOT_LOADER__.register(ye, "GET_FEEDBACK_FORMS_REQUEST", "/Users/vinhlh/Works/vinhlh/demo/demo/assets/js/constants/actionTypes.js"), __REACT_HOT_LOADER__.register(Ee, "GET_FEEDBACK_FORMS_SUCCESS", "/Users/vinhlh/Works/vinhlh/demo/demo/assets/js/constants/actionTypes.js"), __REACT_HOT_LOADER__.register(ge, "GET_FEEDBACK_FORMS_ERROR", "/Users/vinhlh/Works/vinhlh/demo/demo/assets/js/constants/actionTypes.js"), __REACT_HOT_LOADER__.register(Te, "GET_FEEDBACK_FORM_DETAIL", "/Users/vinhlh/Works/vinhlh/demo/demo/assets/js/constants/actionTypes.js"), __REACT_HOT_LOADER__.register(Ce, "GET_FEEDBACK_FORM_DETAIL_REQUEST", "/Users/vinhlh/Works/vinhlh/demo/demo/assets/js/constants/actionTypes.js"), __REACT_HOT_LOADER__.register(be, "
```

How to remove them?
I am using beta 2

thecentric:
Thanks for your comment @tsnieman.  I'm actually using [react-hot-boilerplate](https://github.com/gaearon/react-hot-boilerplate/tree/next) (the `next` branch).  The `webpack.config.js` has no configuration for the module resolution paths, so I've tried adding your suggested code, and also:

    resolveLoader: {
      'fallback': path.resolve(__dirname, "node_modules")
    },

but no luck so far.  I'm also using the module in a custom project with other node modules that load first; they are importing fine, for some reason it's just `react-hot-loader` having issues (and only the `next` branch, and only for my ubuntu machine and an ubuntu VM I tried with; colleagues with OSX are fine).

Strangely we're running an npm build step inside of an ubuntu docker container (during docker image building, not during the running of the container), when building the docker image on my machine the npm build fails (with the same react-hot-loader module import failure), however building the image from a separate container with different base OS (using the [docker base image](https://hub.docker.com/_/docker/) which is Alpine Linux I think) and the npm build is fine.  I'm not really sure what to make of that!

devinrhode2:
How does react-hot-loader relate to redux-devtools?

kwelch:
What are the limitations of hot loading a functional component?

Tatones:
Any timeline for releasing this new version? :) Great work on it by the way!

inithink:
HMR will not work if there was error.
Step:
1. in Counter.js, make error and save.
`return (
      <h2>Counter: {this.state.counter}</h2>
      dummyErrorString
   );`
2. revert it, and change below
`return (
      <h2>Counter1: {this.state.counter}</h2>
);`
3. No update will be found.

manuel-colmenero:
Hey there, I think I discovered a bug in this branch. Is this the correct place to report it? Anyways, here it goes...

Using the generic webpack loader (`react-hot-loader/webpack`) instead of the babel plugin (`react-hot-loader/babel`) causes a weird issue. The loader returns always the same module if the modules' file names are the same (but the paths are different). 

Step 1: Create two modules with different paths but identical filenames:
```
components/test1/index.js
components/test2/index.js
```

Step 2: Load those modules from a third one:
```
import Test1 from "components/test1
import Test2 from "components/test2
```

Both Test1 and Test2 are the same component, I think the first one that was required:
`Test1 === Test2`

Otherwise, congratulations on the new version, it rocks!

ctrlplusb:
@manuel-colmenero are you sure about this?  I have an identical component structure to what you are describing in my projects and haven't had the same issue.  I'm using webpack v2 though... not sure if that makes a difference.  You can have a look at my project's components [here](https://github.com/ctrlplusb/react-universally/tree/master/src/shared/components).

manuel-colmenero:
I am pretty sure about this, I've been "hunting ghosts" for a few hours until I could reproduce it reliably. I stripped everything down so I'm pretty sure it can't be anything else.

A few differences between your boilerplate and mine are:
- I use the latest Webpack version (1.13.1).
- I use `webpack-dev-server` instead of rolling my own dev server.
- My `index.js` files are the modules themselves, they're not empty JS files pointing to the actual module.
- You don't actually use `react-hot-loader/webpack` but the other one, `react-hot-loader/babel`, which also works in my setup. This seems to be the issue.

So to clarify things: `react-hot-loader/babel` (babel plugin, you add it to the babel config) works fine; `react-hot-loader/webpack` (webpack loader, you have to add it to the webpack config along with babel-loader) does *not*.

Right now I can't (I'm at work) but when I have time I'll try to put up a repo reproducing the issue.

AndrewGrachov:
+1 for non working ```react-hot-loader/webpack```
I use pre-babeled files, and it does not reload

peter-mouland:
If anyone is interested, I have implemented an example app, [react-lego](https://github.com/peter-mouland/react-lego) which implements v3. [here are the code changes required to get it working](https://github.com/peter-mouland/react-lego/compare/react-hot-loader).

Faradey27:
@gaearon when you plan to release v3? or maybe to introduce beta version?
is it possible to use hot-reloading for now on large production project without a lot of pain?

peter-mouland:
> I'm not associated with the project so can't talk for the maintainers, these are just my thoughts

@Faradey27 RHL v3.beta3 is available: 
 `npm i -D react-hot-loader@3.0.0-beta.3`

in theory its good to use in a large production website because it should only be used in development... but, It's in Beta, so according to SemVer it may have breaking changes before the final version is released. to find out if more breaking changes are likely before final release we'd have to find out from the guys who are building the project.

I found it quite simple to implement, once I was able to see the changes through the comments, which is why i created the [react-lego](https://github.com/peter-mouland/react-lego) project and this [react-hot-loader branch demonstrating the changes](https://github.com/peter-mouland/react-lego/compare/react-hot-loader).

wmertens:
Also remember, this is dev-only, so the production application is not
impacted. Basically, if it works for you, go for it!

MoOx:
I am seeing this message

```
React Hot Loader: App in /.../web/index.js will not hot reload correctly because index.js uses <App /> during module definition. For hot reloading to work, move App into a separate file and import it from index.js.
```

Here is my index file

```js
// @todo @ flow
/* eslint-disable import/max-dependencies */

import "react-hot-loader/patch" // HMR react patch
import "babel-polyfill"
import "whatwg-fetch"
import { AppRegistry } from "react-native"
import React, { Component } from "react"
import { AppContainer as HotLoaderWrapper } from "react-hot-loader"
import { Provider as StoreProvider } from "react-redux"

import { hideSplashScreen } from "../src/modules/splash-screen"
import AppRoot from "../src/app/AppRoot"
import createStore from "../src/app/redux/store.js"

const store = createStore()

// react-hot-loader need a class
// eslint-disable-next-line react/prefer-stateless-function
class App extends Component {
  render() {
    // require for correct hot loading
    const AppRouter = require("../src/app/AppRouter/web.js").default

    return (
      <HotLoaderWrapper>
        <StoreProvider store={ store }>
          <AppRoot>
            <AppRouter />
          </AppRoot>
        </StoreProvider>
      </HotLoaderWrapper>
    )
  }
}

// copy all assets
const assets = require.context("./assets", true, /.*/)
assets.keys().forEach(assets)

// app placeholder
const rootTag = document.createElement("div")
rootTag.innerHTML = "..."
document.body.appendChild(rootTag)

// start react app, react-native way
AppRegistry.registerComponent("WynsureGroupAdmin", () => App)

AppRegistry.runApplication("WynsureGroupAdmin", { rootTag: rootTag })
hideSplashScreen()

if (module.hot) {
  module.hot.accept("../src/app/AppRouter/web.js", () => {
    AppRegistry.runApplication("WynsureGroupAdmin", { rootTag: rootTag })
  })
}

```

**That said hot loading works perfectly fine**

But this notice was bothering me, so I tried to do what the message said. So I did this

index.js

```js
// @todo @ flow
/* eslint-disable import/max-dependencies */

import "react-hot-loader/patch" // HMR react patch
import "babel-polyfill"
import "whatwg-fetch"
import { AppRegistry } from "react-native"

import { hideSplashScreen } from "../src/modules/splash-screen"

import App from "./App.js"

// copy all assets
const assets = require.context("./assets", true, /.*/)
assets.keys().forEach(assets)

// app placeholder
const rootTag = document.createElement("div")
rootTag.innerHTML = "..."
document.body.appendChild(rootTag)

// start react app, react-native way
AppRegistry.registerComponent("WynsureGroupAdmin", () => App)

AppRegistry.runApplication("WynsureGroupAdmin", { rootTag: rootTag })
hideSplashScreen()

if (module.hot) {
  module.hot.accept("../src/app/AppRouter/web.js", () => {
    AppRegistry.runApplication("WynsureGroupAdmin", { rootTag: rootTag })
  })
}
```

App.js

```js
// @flow

import React, { Component } from "react"
import { AppContainer as HotLoaderWrapper } from "react-hot-loader"
import { Provider as StoreProvider } from "react-redux"

import AppRoot from "../src/app/AppRoot"
import createStore from "../src/app/redux/store.js"

const store = createStore()

// react-hot-loader need a class
// eslint-disable-next-line react/prefer-stateless-function
class App extends Component {
  render() {
    // require for correct hot loading
    const AppRouter = require("../src/app/AppRouter/web.js").default

    return (
      <HotLoaderWrapper>
        <StoreProvider store={ store }>
          <AppRoot>
            <AppRouter />
          </AppRoot>
        </StoreProvider>
      </HotLoaderWrapper>
    )
  }
}

export default App
```

**No message, but hot loading don't work anymore**.

I tried to replace my require().default in App.js by an import, no luck.

I am getting:

```
[HMR] Cannot apply update. Need to do a full reload!
dev-server.js:34 [HMR] Error: Aborted because 106 is not accepted
Update propagation: 106 -> 355
    at hotApply (http://localhost:8080/web.28ffd53129937380cc7e.js:426:30)
    at hotUpdateDownloaded (http://localhost:8080/web.28ffd53129937380cc7e.js:284:13)
    at hotAddUpdateChunk (http://localhost:8080/web.28ffd53129937380cc7e.js:264:13)
    at webpackHotUpdateCallback (http://localhost:8080/web.28ffd53129937380cc7e.js:8:12)
    at http://localhost:8080/0.21264227de06970aa02b.hot-update.js:1:1
```

Then full page loading. What am I doing wrong?

calesce:
@MoOx: dumb question since I haven't used React Native recently: Don't they have their own hot-reloading implementation with their own bundler?

For React Router < 4 I have a simple example for reloading route config [here](https://github.com/calesce/react-hot-loader-examples/blob/439ac1e0623da8d84cbbc851aa9993c29fbaa931/src/index.js). There's a caveat that async routes with `require.ensure` don't work (#288)

It might be worth opening a new issue here (ideally with a repro project), things can get lost in these big threads. :smile:

MoOx:
@calesce sorry I forget to mention it's code for the web, thanks to react-native-web, aliased to react-native via webpack.

calesce:
@MoOx ah ok, yeah I haven't played with that either. I'll take a look later, have you tested that hot-reloading works with a bare-minimum example? I'm guessing `AppRegistry.runApplication` gets turned into `ReactDOM.render`.

MoOx:
Not sure you have read correctly my comment. It's working when everything is in one file (kind of in the example you posted), but I got a warning this will not work and I should explode into 2 files. But with 2 files, it does not work anymore. So this is confusing :)

csillag:
Is there an estimation about when will this come out of beta? Thanks.

calesce:
@csillag see [v3.0 milestone](https://github.com/gaearon/react-hot-loader/milestone/2). No estimates but you can always help :smile:

griffinmichl:
Strangely, I have the page refershing without `module.hot.accept`, but when I add it in, I get the message telling me 'App is up to date' without the changes occurring. Anyone experienced this?

calesce:
@griffinmichl: Yeah I've seen that happen, I'm not sure exactly which cases cause it. Would you mind sharing a minimal project reproducing the issue?

griffinmichl:
Thanks for the offer, but I got it worked out. Didn't realize I had to re-require `<App />` inside module.hot.accept's callback. I must have been looking at a bad example or old branch.

griffinmichl:
Something I found really useful for getting this up and running after a lot of struggle (especially with koa 2, which has less up to date webpack middleware): just use the webpack-dev-server cli and configure `devServer` in your webpack config to proxy API and asset requests to your app server. Ended up being much easier to implement and you don't end up with a bunch of dev vs prod logic in your server.

peter-mouland:
@griffinmichl i'd be interested in seeing it with Koa v2, have you an example app for a demo?

akoskm:
@griffinmichl yes, I've seen that happening too. The WDS/HMR messages appeared in the console but the App remained the same. I never removed `module.hot.accept`, but instead of

```
module.hot.accept('./components/main', () => { // .. });
```

I switched to

```
module.hot.accept(() => { // .. });
```

and it suddenly started working. Here is the relevant commit from my boilerplate project: https://github.com/akoskm/react-hot-boilerplate/commit/9938a629b7825523e8d825b42523a0d9fa0e7d53

Since it started working without the `'./components/main'` it also works when that parameter is present. 😲

ntucker:
Whenever I change a module, the console logs say it's updated, but nothing changes. I did some tracking on module updates:

- Changed module is being executed
- Component render function is being run
- However, the module code is still old (changing the console.log message doesn't get reflected).

Is webpack somehow not actually updating the code, but just re-executing the existing modules? (I'm using webpack 1.14.0)

dobryanskyy:
@ntucker were you able to solve the issue? I'm experiencing quite similar one

ntucker:
@dobryanskyy nope, still broken.

rolandfung:
I have multiple applications in the same repo, so I use a wrapper function that takes in the root element, store and controller as the parameters (I'm using Cerebral.js). Changes in application files are being detected but I get full reload each time in the browser. Is my usage of this wrapper causing the problem? I am on RHL3 beta 6 and webpack 1.13.2.

```
let getAppWrapper = (App, controller, store) => {
    @root(store)
    class AppWrapper extends React.Component {
        render() {
            return (
                <Container controller={controller}>
                    <MuiThemeProvider>
                        <App {...this.props}>
                            {this.props.children}
                        </App>
                    </MuiThemeProvider>
                </Container>
            );
        }
    }
    return AppWrapper;
};
```

ismay:
@rolandfung I'm experiencing the same problem (full reload on changes), with hot-loader beta6 and webpack 2.2.0:

```javascript
// abstract render for hot reloading
const renderWithHotReload = AppComponent => {
  render(
    <AppContainer>
      <AppComponent store={store} />
    </AppContainer>,
    document.getElementById('app')
  )
}

// initial render
renderWithHotReload(App)

// enable hot reloading, will be stripped in production
if (process.env.NODE_ENV !== 'production' && module.hot) {
  module.hot.accept('./components/app', () => {
    const NextApp = require('./components/app').App
    renderWithHotReload(NextApp)
  })
  module.hot.accept('./rootReducer', () => {
    const nextRootReducer = require('./rootReducer').default
    store.replaceReducer(nextRootReducer)
  })
}
```

Code [here](https://github.com/ismay/ismaywolff.nl/blob/develop/client/index.jsx#L19). Apologies if this isn't the place to post this (but thought it would be valuable in case it's a bug).

SimenB:
Woo! :tada:

wkwiatek:
Just to keep you up-to-date. 

We're now finalizing docs, and have 3 more bugs to solve that are blockers for final release. All of these are on the behalf of `react-proxy`.

We agreed to merge `next` to the master as some people had problem with starting with React Hot Loader, and v1 is now stale.

SimenB:
React-proxy hasn't seen a commit in 11 months, will it be getting some love?
And could you link to the outstanding issues?

wkwiatek:
Here are all of them: https://github.com/gaearon/react-hot-loader/milestone/2

I'm going to focus on the `react-proxy` right after RHL v3 docs are done.

PR #250, opened 2016-04-19
Makes App Container take children, closes #244.

### LINKED ISSUE #244: Consider making <AppContainer> take children
Currently it looks like this:

```js
<AppContainer
  component={Root}
  props={{ store }} /* depending on what you want to pass */
/>
```

As suggested by @halt-hammerzeit in https://github.com/gaearon/redux-devtools/commit/64f58b7010a1b2a71ad16716eb37ac1031f93915#commitcomment-17142917 we might want to make it accept children instead:

```js
<AppContainer>
  <Root store={store} />
</AppContainer>
```

I think I’m on board with this. A pull request is welcome!

lolJS:
Tested locally seems to hot reload stuff.

### DIFF
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -82,8 +82,7 @@ class AppContainer extends Component {
 
     if (this.props.component) {
       return <this.props.component {...this.props.props} />;
-    }
-    else {
+    } else {
       return React.Children.only(this.props.children);
     }
   }
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -49,7 +49,7 @@ class AppContainer extends Component {
     // Hot reload has finished.
     // Force-update the whole tree, including
     // components that refuse to update.
-    if (prevProps.children !== this.props.children) {
+    if (prevProps.children.type !== this.props.children.type) {
       deepForceUpdate(this);
     }
   }
@@ -69,10 +69,22 @@ class AppContainer extends Component {
     if (error) {
       return <this.props.errorReporter error={error} />;
     } else {
-      return React.cloneElement(this.props.children);
+      return React.Children.only(this.props.children);
     }
   }
 }
+
+AppContainer.propTypes = {
+  children: function (props, propName, componentName, location, propFullName) {
+    if (typeof props[propName].type !== 'function') {
+      return new Error(`Invalid prop ${propFullName} supplied to ${componentName}. Expected a React Component!`);
+    }
+
+    if (React.Children.count(props[propName]) > 1) {
+      return new Error(`Invalid prop ${propFullName} supplied to ${componentName}. Expected a single React Component!`);
+    }
+  }
+};
 
 AppContainer.defaultProps = {
   errorReporter: Redbox
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -38,13 +38,7 @@ class AppContainer extends Component {
   componentWillReceiveProps(nextProps) {
     // Hot reload is happening.
     // Retry rendering!
-    if (nextProps.component !== this.props.component) {
-      this.setState({
-        error: null
-      });
-    }
-
-    if (nextProps.children.type !== this.props.children.type) {
+    if (this.hasChildTypeChanged(nextProps)) {
       this.setState({
         error: null
       });
@@ -55,11 +49,7 @@ class AppContainer extends Component {
     // Hot reload has finished.
     // Force-update the whole tree, including
     // components that refuse to update.
-    if (prevProps.component !== this.props.component) {
-      deepForceUpdate(this);
-    }
-
-    if (prevProps.children.type !== this.props.children.type) {
+    if (this.hasChildTypeChanged(prevProps)) {
       deepForceUpdate(this);
     }
   }
@@ -72,6 +62,19 @@ class AppContainer extends Component {
     this.setState({
       error: error
     });
+  }
+
+  hasChildTypeChanged(props) {
+    // Backwards compat
+    if (props.component !== this.props.component) {
+      return true;
+    }
+
+    if (props.children.type !== this.props.children.type) {
+      return true;
+    }
+
+    return false;
   }
 
   render() {
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -38,6 +38,12 @@ class AppContainer extends Component {
   componentWillReceiveProps(nextProps) {
     // Hot reload is happening.
     // Retry rendering!
+    if (nextProps.component !== this.props.component) {
+      this.setState({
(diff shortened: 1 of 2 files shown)


src/AppContainer.dev.js
```diff
@@ -69,7 +69,7 @@ class AppContainer extends Component {
     if (error) {
       return <this.props.errorReporter error={error} />;
     } else {
-      return <this.props.component {...this.props.props} />;
+      return React.cloneElement(this.props.children);
```
gaearon: Let’s return `React.Children.only(this.props.children)` instead?
This way we enforce that there’s a single child.
`React.cloneElement()` isn’t necessary here.

src/AppContainer.dev.js
```diff
@@ -49,7 +49,7 @@ class AppContainer extends Component {
     // Hot reload has finished.
     // Force-update the whole tree, including
     // components that refuse to update.
-    if (prevProps.component !== this.props.component) {
+    if (prevProps.children !== this.props.children) {
```
gaearon: I think we want to compare `props.children.type` and force-update if it changed. Otherwise, if the user re-renders from the top for other reasons than a hot reload, we are wasting force updates. Sure, the user probably wouldn’t do that, but it’s better to double check.

gaearon:
Thanks for the PR! Left a few comments inline. I think that we want to add a custom `propType` for `children` that verifies that this is a single element, and that its `type` is a function. We shouldn’t allow passing `<div><Stuff /></div>`, for example.

lolJS:
@gaearon Updated the PR per your comments. I'm not sure if the messaging on the warnings I wrote are clear enough.

src/AppContainer.dev.js
```diff
@@ -38,7 +38,7 @@ class AppContainer extends Component {
   componentWillReceiveProps(nextProps) {
     // Hot reload is happening.
     // Retry rendering!
-    if (nextProps.component !== this.props.component) {
+    if (nextProps.children !== this.props.children) {
```
gaearon: Should also compare `type` here.

src/AppContainer.dev.js
```diff
@@ -69,11 +69,23 @@ class AppContainer extends Component {
...
-      return <this.props.component {...this.props.props} />;
+      return React.Children.only(this.props.children);
     }
   }
 }
 
+AppContainer.propTypes = {
+  children: function (props, propName, componentName, location, propFullName) {
+    if (typeof props[propName].type !== 'function') {
+      return new Error(`Invalid prop ${propFullName} supplied to ${componentName}. Expected a React Component!`);
```
gaearon: `Expected a single React element with your app’s root component, e.g. <App />.`

src/AppContainer.dev.js
```diff
@@ -69,11 +69,23 @@ class AppContainer extends Component {
...
+    }
+
+    if (React.Children.count(props[propName]) > 1) {
+      return new Error(`Invalid prop ${propFullName} supplied to ${componentName}. Expected a single React Component!`);
```
gaearon: I’d usage the same message here.

alexanderchr:
Might want to update `AppContainer.prod.js` as well

lolJS:
Good call. I'll get on that :)

src/AppContainer.dev.js
```diff
@@ -69,11 +69,23 @@ class AppContainer extends Component {
...
 }
 
+AppContainer.propTypes = {
+  children: function (props, propName, componentName, location, propFullName) {
+    if (typeof props[propName].type !== 'function') {
+      return new Error(`Invalid prop ${propFullName} supplied to ${componentName}. Expected a single React element with your app’s root component, e.g. <App />.`);
+    }
+
+    if (React.Children.count(props[propName]) > 1) {
+      return new Error(`Invalid prop ${propFullName} supplied to ${componentName}. Expected a single React Component!`);
```
gaearon: Error message needs to be updated here too. It’s fine if it’s exactly the same one.

gaearon:
Looks great!

Can you please add support for the old way of doing things? The propType warning is enough to get people to migrate to the new way, but I’d rather be able to release it in the next alpha without bumping to beta. I’ll remove the compat code in a beta.

gaearon:
The new code doesn’t.

src/AppContainer.dev.js
```diff
@@ -38,6 +38,12 @@ class AppContainer extends Component {
   componentWillReceiveProps(nextProps) {
     // Hot reload is happening.
     // Retry rendering!
+    if (nextProps.component !== this.props.component) {
```
lolJS: @gaearon Thought of this after I committed...should we check type here too?

src/AppContainer.prod.js
```diff
@@ -3,7 +3,7 @@ const { Component } = React;
 
 class AppContainer extends Component {
   render() {
-    return <this.props.component {...this.props.props} />;
+    return React.Children.only(this.props.children);
```
gaearon: We need backward compat here too.

src/AppContainer.dev.js
```diff
@@ -68,12 +78,29 @@ class AppContainer extends Component {
     const { error } = this.state;
     if (error) {
       return <this.props.errorReporter error={error} />;
-    } else {
+    }
+
+    if (this.props.component) {
       return <this.props.component {...this.props.props} />;
     }
+    else {
```
gaearon: Let's move else to the previous line.

src/AppContainer.dev.js
```diff
@@ -43,6 +43,12 @@ class AppContainer extends Component {
         error: null
       });
     }
+
+    if (nextProps.children.type !== this.props.children.type) {
```
gaearon: Let's extract this to `hasChildTypeChanged` and use it everywhere we compare children. It would include backward compat logic which we would later remove.

levithomason:
Looks ready to go, bump.

gaearon:
Yeah, this looks great. Will release today or tomorrow.

gaearon:
Backward compat doesn’t work here.

1. It crashes inside the proptype check because it assumes `children.type` exists
2. After fixing this, it crashes in `hasChildTypeChanged` for the same reason

I’ll fix this but I’d appreciate more thorough manual testing next time 😉 .

gaearon:
Out in `react-hot-loader@3.0.0-beta.0`.

levithomason:
Awesome, will give this a whirl soon.

alexanderchr:
Works flawlessly. Thanks.

lolJS:
Awesome! Sorry didn't think to test the backwards compat since I had already converted my test code base to use the children. I'll be more thorough next time :+1:

levithomason:
Confirmed, looks good so far here 😸

PR #252, opened 2016-04-21
Fix hack for React Router to handle also IndexRoute

zywyz:


gaearon:
Are there any other kinds of routes in RR?

alexanderchr:
Yeah, there are plain routes. Think they will be hard to support like this though as there are no `Route` elements at all when using them.

gaearon:
We should probably change the hack to activate for `Router` component and traverse the `children` and `routes` props, making sure that we support both APIs (and that we do this defensively so if there’s an API change, we don’t fail hard). https://github.com/gaearon/react-hot-boilerplate/pull/61#issuecomment-211504531 gives an idea of how that looks like for React element routes, but plain object routes work the same, they just use arrays and have different field names.

@zywyz Would you be interested in looking into this? I’m fine merging the current PR but a more comprehensive hack might be better.

gaearon:
👍 

I’ll cut a release in a day or two.

mondaychen:
@gaearon Can you release a new alpha version for this? Thanks!

gaearon:
I’ll try to do this tomorrow, with a more comprehensive fix from #253.

PR #253, opened 2016-04-21
Allow routes from routes prop

alexanderchr:
As discussed in #252.

### DIFF
--- a/src/fixupReactRouter.js
+++ b/src/fixupReactRouter.js
@@ -49,4 +49,4 @@ function fixupReactRouter(props, forceUpdateFunction) {
   });
 }
 
-module.exports = fixupReactRouter;
+module.exports = { isReactRouterish, fixupReactRouter };
--- a/src/fixupReactRouter.js
+++ b/src/fixupReactRouter.js
@@ -49,4 +49,4 @@ function fixupReactRouter(props, forceUpdateFunction) {
   });
 }
 
-module.exports = hackRouter;
+module.exports = fixupReactRouter;
--- a/src/hackRouter.js
+++ b/src/hackRouter.js
@@ -1,5 +1,3 @@
-const resolveType = require('./resolveType').default;
-
 // This is lame but let's focus on shipping.
 // https://github.com/gaearon/react-hot-loader/issues/249
 function isReactRouterish(type) {
@@ -9,12 +7,12 @@ function isReactRouterish(type) {
   );
 }
 
-function forceUpdateComponentsOfRouteAndChildRoutes(route) {
+function forceUpdateComponentsOfRouteAndChildRoutes(route, forceUpdateFunction) {
   // TODO: check whether it is possible to also handle the `getComponent` case here
   if (route.component && typeof(route.component) === 'function') {
     // Side effect 😱
     // Force proxies to update since React Router ignores new props.
-    resolveType(route.component);
+    forceUpdateFunction(route.component);
   }
 
   if (route.components) {
@@ -24,7 +22,7 @@ function forceUpdateComponentsOfRouteAndChildRoutes(route) {
       const component = route.components[key];
 
       if (typeof(component) === 'function') {
-        resolveType(component);
+        forceUpdateFunction(component);
       }
     }
   }
@@ -38,13 +36,13 @@ function forceUpdateComponentsOfRouteAndChildRoutes(route) {
   for (const childRoute of childRoutes) {
     // When using `Route` element routes the relevant objects will be under props
     // and when using plain routes directly on the route
-    forceUpdateComponentsOfRouteAndChildRoutes(childRoute.props || childRoute);
+    forceUpdateComponentsOfRouteAndChildRoutes(childRoute.props || childRoute, forceUpdateFunction);
   }
 }
 
-export default function hackRouter(type, props) {
+export default function hackRouter(type, props, forceUpdateFunction) {
   if (isReactRouterish(type) && props) {
-    forceUpdateComponentsOfRouteAndChildRoutes(props);
+    forceUpdateComponentsOfRouteAndChildRoutes(props, forceUpdateFunction);
   }
 }
 
--- a/src/hackRouter.js
+++ b/src/hackRouter.js
@@ -1,0 +1,51 @@
+const resolveType = require('./resolveType').default;
+
+// This is lame but let's focus on shipping.
+// https://github.com/gaearon/react-hot-loader/issues/249
+function isReactRouterish(type) {
+  return type && (
+    type.displayName === 'Router' ||
+    type.name === 'Router' // In case Ryan and Michael embrace ES6 classes
+  );
+}
+
+function forceUpdateComponentsOfRouteAndChildRoutes(route) {
+  // TODO: check whether it is possible to also handle the `getComponent` case here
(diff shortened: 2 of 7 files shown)


alexanderchr:
Might actually have to do this for all child routes as well, will test.

src/patch.dev.js
```diff
@@ -77,26 +77,43 @@ if (React.createElement.isPatchedByReactHotLoader) {
...
-    props &&
-    typeof props.component === 'function'
-  ) {
-    // Side effect 😱
-    // Force proxies to update since React Router ignores new props.
-    resolveType(props.component);
+  if (isReactRouterish(type) && props) {
+    // A router's routes may be given as children or through
+    // the `routes` prop.
+    var topRoutes = [].concat(props.routes, props.children).filter(Boolean);
```
gaearon: Let’s move this initial loop into `forceUpdateRouteAndChildRouteComponents` as well.

src/patch.dev.js
```diff
@@ -77,26 +77,43 @@ if (React.createElement.isPatchedByReactHotLoader) {
...
+    // Side effect 😱
+    // Force proxies to update since React Router ignores new props.
+    resolveType(route.component);
+  }
+
+  // Child routes will be in `props.children` when defining routes using react
+  // elements and in `childRoutes` when using plain routes
+  var childRoutes = [].concat(route.childRoutes, route.props && route.props.children).filter(Boolean);
+
+  for (var _len = childRoutes.length, _key = 0; _key < _len; _key++) {
```
gaearon: Let’s just use `for (var i = 0; i < childRoutes.length; i++)`?

src/patch.dev.js
alexanderchr: Sure.

src/patch.dev.js
```diff
@@ -77,26 +77,42 @@ if (React.createElement.isPatchedByReactHotLoader) {
...
+  // elements and in `routes` or `childRoutes` when using plain routes
+  var childRoutes = []
+    .concat(
+      route.routes,
+      route.childRoutes,
+      route.props && route.props.children)
+    .filter(Boolean);
+
+  for (var i = 0; i < childRoutes.length; i++) {
+    var childRoute = childRoutes[_key];
```
gaearon: Still using `_key` here

alexanderchr: Yeah, nice catch. It seems like it worked because babel were transpiling it to _key anyways.

src/patch.dev.js
```diff
@@ -77,26 +77,42 @@ if (React.createElement.isPatchedByReactHotLoader) {
...
+  // TODO: check whether it is possible to also handle the `getComponent` case here
+  if (route.component && typeof route.component === 'function') {
+    // Side effect 😱
+    // Force proxies to update since React Router ignores new props.
+    resolveType(route.component);
+  }
+
+  // Child routes will be in `props.children` when defining routes using react
+  // elements and in `routes` or `childRoutes` when using plain routes
+  var childRoutes = []
```
gaearon: Minor nit: let’s use `const` where possible, `let` otherwise.

alexanderchr: Sure.

src/patch.dev.js
```diff
@@ -77,26 +77,43 @@ if (React.createElement.isPatchedByReactHotLoader) {
...
-    type.name === 'IndexRoute'
+    type.displayName === 'Router' ||
+    type.name === 'Router' // In case Ryan and Michael embrace ES6 classes
   );
 }
 
+function forceUpdateComponentOfRouteAndChildRoutes(route) {
+  // TODO: check whether it is possible to also handle the `getComponent` case here
+  const components = []
+    .concat(route.component, route.components && Object.values(route.components))
```
gaearon: `Object.values` may not work in older browsers, we shoot for wide compatibility.

src/patch.dev.js
```diff
@@ -77,26 +77,43 @@ if (React.createElement.isPatchedByReactHotLoader) {
...
+    .filter(c => typeof(c) === 'function');
+
+  for(const component of components) {
```
gaearon: Nit: space after `for`.

src/patch.dev.js
alexanderchr: Right, changed it a bit.

gaearon:
Thanks, this looks good. I have a few more style nits but I’m happy to fix those myself. I need to add a linter anyway.

Would you be open to considering adding some tests for this? I have a basic Mocha setup for the Babel plugin, and it seems like this is another example of something we want to be tested.

src/hackRouter.js
```diff
@@ -0,0 +1,48 @@
...
+function objectValues(obj) {
+  let values = [];
+  for (const key in obj) {
+    if (obj.hasOwnProperty(key)) {
+      values.push(obj[key]);
+    }
+  }
+  return values;
+}
+
```
alexanderchr: Might want to use a real polyfill instead

alexanderchr:
Sure, added a few tests.

Also had some midnight inspiration and rewrote all of it. I included these changes in the last commit, so feel free to discard them if you think the old way was better. Otherwise you can just squash them into the first refactor commit. Tests should hopefully be passing either way.

src/hackRouter.js
gaearon: Maybe we’d just use `Object.keys(obj).map(key => obj[key])`?

gaearon:
This looks great! `hackRouter` sounds a little intimidating, let’s rename it to `addReactRouterFix()` or something similar?

alexanderchr:
Sure!

src/hackRouter.js
alexanderchr: Yeah, that's it. Thanks.

src/fixupReactRouter.js
```diff
@@ -0,0 +1,52 @@
...
+    .filter(c => typeof(c) === 'function');
+}
+
+function fixupReactRouter(props, forceUpdateFunction) {
+  // Ideally we want to teach React Router to receive children.
+  // We're not in a perfect world, and a dirty workaround works for now.
+  // https://github.com/reactjs/react-router/issues/2182
+  const routes = props.routes || ensureArray(props.children);
+  const components = componentsToForceUpdate(routes);
+  for (const component of components) {
```
gaearon: [`for of` requires `Symbol`s](http://babeljs.io/repl/#?evaluate=true&presets=es2015%2Creact&experimental=false&loose=false&spec=false&code=for%20(var%20x%20of%20y)%20%7B%0A%20%20%0A%7D), let’s use `components.forEach`

src/fixupReactRouter.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function isReactRouterish(type) {
+  return type && (
+    type.displayName === 'Router' ||
+    type.name === 'Router' // In case Ryan and Michael embrace ES6 classes
+  );
+}
+
+function componentsToForceUpdate(routes) {
```
gaearon: Let’s rename to `extractComponents`. We’re not exactly force updating them anyway.

gaearon:
Out in `3.0.0-alpha.13`, thank you!

alexanderchr:
Cool. Thanks for reviewing!

PR #254, opened 2016-04-22
[docs] [WIP] Migration Guide, Features

hedgerh:
This is non-website doc stuff.  Made a first pass at a Migration to 3.0 guide and added stuff to main README.

I also think the list of starter kits/boilerplates doesn't really make sense in the scope of this project, and should be taken out, with the exception of `react-hot-boilerplate`.  A decent few of them were outdated/abandoned anyway.  I can add them back if you don't agree, though.


gadicc:
Merged the commits into `gh-pages-next` (tracking in #257) so we can collaborate on it.

Issue #243, opened 2016-04-18
Document React Hot Loader 3


gaearon:
It features two modes (`react-hot-loader/babel` and `react-hot-loader/webpack`), doesn’t hide HMR API (you have to add `module.hot.accept` call yourself) and has a somewhat different API and setup process.

There’s info in https://github.com/gaearon/react-hot-boilerplate/pull/61 and https://github.com/gaearon/redux-devtools/commit/64f58b7010a1b2a71ad16716eb37ac1031f93915 but we should create good step-by-step docs and update the website too.

This will take a while because it’s not a priority for me. If you want to contribute, you are very welcome!
Obviously we can’t release 3.0 without any docs.

gaearon:
To give some sense of perspective, if nobody contributes, the stable 3.0 might get out as late as June because I really don’t have the bandwidth to work on a good doc right now. It really is the biggest blocker to getting this version out!

pablopaul:
@gaearon What kind of documentation is needed? Maybe you can give a rough documentation outline or so…

gaearon:
We need to update [this intro](http://gaearon.github.io/react-hot-loader/getstarted/).

The new intro would be slightly longer because there is a bit more setup involved.
I would split it into several sections:

1. Enabling HMR
  * Create a development Webpack config separate from production one
  * Add `HotModuleReplacementPlugin` to development Webpack config
  * If you only render on the client, consider using `WebpackDevServer`
    * Easier to set up
    * Enable `hot: true` and add its entry points
  * If you use server rendering, consider using Express server + `webpack-dev-middleware`
    * More work but also more control
    * Show how to add `webpack-dev-middleware` and its entry point
2. Using HMR to replace the root component
  * Show how to add `module.hot` hook to re-render your root component
  * (This will not preserve state but is good enough for some cases!)
3. Adding React Hot Loader to preserve state
  * If you use Babel:
    * Add `react-hot-loader/babel` to `plugins`
  * If you don’t use Babel:
    * Add `react-hot-loader/webpack` to `loaders`
  * Add `react-hot-loader/patch` as the first entry point
  * Change your `index.js` to render `<AppContainer>` instead of your root component
  * That’s it!

hedgerh:
I can help out with the docs.  I've been reading up and gathering as much info as I can.  I'm about to go through both the old and 3.0 source to try and figure out what's going on under the hood.  I'll see if I can get the ball rolling on the Intro, as well.

[...]

PR #258, opened 2016-04-26
Adding clarification to RHL3 docs + JSX highlight in code blocks

tsaiDavid:
@gadicc @hedgerh @gaearon - Went ahead and added the changes I mentioned along with some other little touches.

- Clarification and showing the JSX syntax with the AppContainer component wrapping RootContainer - hopefully this is more beginner friendly and less confusing since v3 will require a bit more config.
- Switched from `js` to `jsx` for the code blocks, if I'm not mistaken it's GitHub markdown friendly 👍 
- Provided super basic start for us to show usage within webpack loader (just to get the ball rolling)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -68,7 +68,7 @@ When the HMR runtime receives an updated module, it first checks to see if the m
 
 If your client entry point looks like this:
 
-```js
+```jsx
 import React from 'react';
 import { render } from 'react-dom';
 import RootContainer from './containers/rootContainer.js';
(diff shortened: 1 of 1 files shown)


tsaiDavid:
Just want to point out I was also able to update the shell command example, prefixed it with the `$` symbol that we typically see, indicating that it is a command to be used from one's terminal.

An example was also not in consistent ES6 as in our other examples, so I updated it.

tsaiDavid:
@gadicc #257 #243 - apologies for all the comments, just want to make sure I'm referencing the issue correctly, please let me know if there's anything else I can do.

gadicc:
this is awesome, @tsaiDavid, thanks!  all very welcome improvements.  just one issue which I'll comment on in the diff and then I'm happy to merge.  yeah, for future it's enough to just put a #257 in there somewhere and we'll get a notification (I think :)).

docs/README.md
```diff
@@ -20,7 +20,8 @@ First we had React Hot Loader, which was superseded by React Transform.  Each ti
 *   Works with or without Babel
 *   Everything you need in a single repo
 
-For more information on the evolution of approaches taken, see [](https://medium.com/@dan_abramov/hot-reloading-in-react-1140438583bf)https://medium.com/@dan_abramov/hot-reloading-in-react-1140438583bf.
+For more information on the evolution of approaches taken
```
gadicc: just this snuck through :)

gadicc:
@tsaiDavid I'm going to fix that line on my side, these were actually intended for the gh-pages (sorry that's my bad) so need to move these there and then I'll ping you again.  Thanks!

PR #269, opened 2016-04-30
added failing AppContainer test for force update when passing component as children

nfcampos:
see the failing test.
it's because this line https://github.com/gaearon/react-hot-loader/blob/next/src/AppContainer.dev.js#L77 doesn't work as expected — since the component was passed as children, by the time it appears in AppContainer's lifecycle methods it has already gone through React.createElement, therefore it has already been replaced the proxy, thus (unless react-proxy malfunctions) `children.type` will always be equal to that of the previous props.

I was actually not out to find this bug. I was experimenting with writing tests for AppContainer, I guess the fact I've found this bug kind of makes the case for the tests ;)

Should I write some more tests for AppContainer, specifically does it make sense to write tests for functional components, HOCs, etc. making sure they all properly update? Or is that all covered by the tests for `react-proxy`?

### DIFF
--- a/test/AppContainer/AppContainer.dev.js
+++ b/test/AppContainer/AppContainer.dev.js
@@ -1,0 +1,160 @@
+import './setup'
+import React, {Component} from 'react'
+import expect, {createSpy} from 'expect'
+import {mount} from 'enzyme'
+
+import AppContainer from '../../src/AppContainer.dev'
+
+const tag = (comp, name) => comp.__source = { fileName: name, localName: name }
+
+describe('<AppContainer />', () => {
+  describe('when passed children', () => {
+    it('renders it', () => {
+      const spy = createSpy()
+      class App extends Component {
+        render() {
+          spy()
+          return <div>hey</div>
+        }
+      }
+      tag(App, 'App')
+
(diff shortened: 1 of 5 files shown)


test/AppContainer/AppContainer.dev.js
```diff
@@ -0,0 +1,160 @@
...
+          return <div>hey</div>
+        }
+      }
+      tag(App, 'App')
+
+      const wrapper = mount(<AppContainer><App /></AppContainer>)
+      expect(spy.calls.length).toBe(1)
+
+      {
+        class App extends React.Component {
```
nfcampos: yay for block-scoped class declarations :)

nfcampos:
(I don't see any way to allow providing the component as `children` and not force update every time `componentDidUpdate` is called, but I might be wrong.)

gaearon:
Thanks a lot. Let's change it to force update every time then, it's fine in development. Yeah, more tests would be welcome. While React Proxy is well tested we want at least a couple "integration tests" making sure we use it correctly.

gaearon:
Added you to collaborators. Please feel free to fix the bug on the next branch directly.

nfcampos:
Thanks. I will.

PR #270, opened 2016-04-30
Add warning about using functional root component

jiayihu:
When the root component is written as stateless plan function, instead of `React.Component`, the modules won't reload and you always get the `The following modules couldn't be hot updated: (They would need a full reload!)` warning.
This can be easily replicated changing the App.js of the [boilerplate](https://github.com/gaearon/react-hot-boilerplate) into:
```javascript
import React, { Component } from 'react';

export default function App() {
  return (
    <h1>Hello, world.</h1>
  );
}
```

I don't know if it is already fixed in the new version or if it was already known as problem, but I lost some hour before discovering it on my own so I think it would help adding it to the docs to save someone else's time 😬  
(My root component was written as pure function because of ESLint with Airbnb's config, which warns about writing stateless components as classes.)

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -51,7 +51,7 @@ module.exports = {
 
 If you used WebpackDevServer CLI mode and after switching to Node it crashes with `Error: Invalid path ''`, you probably didn't have `path` specified in `output` at all. You can just put `path: __dirname` there, as it won't matter for development config.
 
-### Module not found: Error: Cannot resolve module 'react-hot' 
+### Module not found: Error: Cannot resolve module 'react-hot'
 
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks! This is indeed a known problem in 1.x but it’s solved completely in [the upcoming 3.x](https://github.com/gaearon/react-hot-boilerplate/pull/61). Would you please amend the note to mention that?

jiayihu:
Yes, sure! I did it right now 👍

gaearon:
👍

kkak10:
@gaearon I have one more question. React.PureComponent too not reloading now. react-hot-reload 3 is solve PureComponent hot reload?

gtabmx:
@kkak10, Just out of curiosity, were you able to get hot loading working with PureComponents? I'd rather not convert to Components.

kkak10:
@gtabmx after, I'm not try react-hot-reload 3 in PureComponent hot reload. I'm doing another Project now...

Issue #249, opened 2016-04-18
Get this working properly with React Router


gaearon:
There is a hacky workaround in https://github.com/gaearon/react-hot-boilerplate/pull/61#issuecomment-211504531 but it’s way too hacky. We should either implement this in https://github.com/reactjs/react-router/issues/2182, or include some sort of hacky workaround right in this project so at least it’s on us. Special casing React Router doesn’t sound very great but it’s hugely popular and we can’t just not support it.
- [x] Document hacky workaround
- [x] Add hacky workaround so it works out of the box (`3.0.0-alpha.12`)
- [x] Figure out proper support out of the box
- [ ] Check how it works with HOCs (http://stackoverflow.com/q/36617240/458193)
- [ ] Check Relay integration (https://github.com/gaearon/react-hot-boilerplate/pull/61#issuecomment-211539886)

glenjamin:
I ran into this today when using react router properly for the first time, and was surprised that no-one had tackled this before.

How do all those people work with a Router that won't hot reload without getting really frustrated?

Ideally, the proper fix for this should be in resolving https://github.com/reactjs/react-router/issues/2182 - bcause the warning here is just cheeky: https://github.com/reactjs/react-router/blob/master/modules/Router.js#L131 "Sorry, this looks like a declarative component but it's actually an imperative API call and we're doing to ignore your new routes!"

I'll take my further comments / help over to that thread now :)

gaearon:
Previous versions of React Hot Loader or React Transform used to wrap class in the proxy during its definition time, so this wasn’t much of an issue—the update would just never propagate up to the router. The current approach has many more upsides (e.g. works with functional components) but now it needs all components in chain to update the React way, so the Router behavior becomes problematic.

gaearon:
Thanks for looking at it!

glenjamin:
I just had an idea I wanted to write down, not thought through heavily yet.

What if instead of trying to patch `<Router>`, you patch `<Route>` to apply the proxy code.

Something along these lines:

```jsx
require('react-hot-loader/patch-router')(Route);

// ...
React.createElement = function(type, props, ..children) { 
  // as well as other stuff
  if (type == Router && props.component) props.component = proxy(props.component)
  // ...
}
```

This sort of approach would ensure that the component is proxied whenever the routes list is "rendered" even though those `<Route>` children are never actually applied to `<Router>`

[...]

PR #271, opened 2016-05-01
(WIP) Get rid of _source

### LINKED ISSUE #266: React Hot Loader 3.0 might prevent views updates
Views are no more updated when we connect a container to an externally defined simple component. See this commit loicfrering/redux-devtools@d387bc9b6dd517e91f6d4c1008490cd4b55720a9 for reference.

When I start the app, actions are correctly dispatched (as we can see in the devtools) but the views are not updating anymore.

And when I remove `react-hot-loader/patch` from `webpack.config.js`, everything is working as expected.

Tested with react-hot-loader 3.0.0-alpha.13. I'll try to investigate more over the weekend, thanks!

gaearon:
Should fix #266 and #249.


gaearon:
Wrong branch.

PR #272, opened 2016-05-01
(WIP) Get rid of _source

### LINKED ISSUE #266: React Hot Loader 3.0 might prevent views updates

gaearon:
Should fix #266 and #249.

TODO:

- [x] Fix Webpack loader as well
- [x] Only rely on WeakMap if it's available
- [x] Use `global` npm package instead of `window`
- [x] Revive the warning about calling `ReactDOM.render()` in the same file
- [x] Test edge cases
- [x] Test React Router support (manually)


nfcampos:
You might want to test this with the kind of tests I have in #269

src/resolveType.js
```diff
@@ -1,72 +1,36 @@
...
-    if (!warnedAboutTypes[id]) {
-      warnedAboutTypes[id] = true;
-      console.error(
-        `React Hot Loader: ${source.localName} from ${source.fileName} will not ` +
-        `hot reload correctly because it contains an imperative call like ` +
-        `ReactDOM.render() in the same file. Split ${source.localName} into a ` +
-        `separate file for hot reloading to work.`
-      );
-    }
+  var id = idsByType.get(type);
```
nfcampos: I think this will update the proxy even if createElement is called with stale type only if we instead keep a typesById map and do `var latestType = typesById.get(id)` after this line and use this latestType to update the proxy. Otherwise it doesn't solve the router problem

gaearon: The reason stale types were used is because Router cached old `type`s and called `createElement` with them. Now the only “entry point” to updating a proxy is an explicit `__REACT_HOT_LOADER__` call. So router can’t revive old types.

nfcampos: Good point.

PR #273, opened 2016-05-01
More tests for AppContainer

nfcampos:
I'll keep this open while I add some more tests.
This also includes moving `deepForceUpdate(this)` to `componentWillReceiveProps`.
I've also removed the tests for receiving the component as prop seeing as that is deprecated.

Todo:
- [x] test with HOC-wrapped component as root


gaearon:
Really nice job with the tests.

Issue #276, opened 2016-05-02
version 3 loses component local state


jlongster:
I have an accordion that has local state that tracks which panes are open. When I update one of the panes, the open panes collapse because (I'm assuming) the accordion's state is being set back the default state.

There are many moving pieces so I'm not sure what to point out. I'm using webpack-dev-middleware, webpack-hot-middleware, and my config looks like this:

```js
const config = Object.assign({}, projectConfig, {
  entry: [
    "webpack-hot-middleware/client?path=/__webpack_hmr&timeout=2000&quiet=true",
    "react-hot-loader/patch",
    path.join(__dirname, "../public/js/main.js")
  ]
});

config.plugins = config.plugins.concat([
  new webpack.HotModuleReplacementPlugin(),
  new webpack.NoErrorsPlugin()
]);

config.module.loaders = config.module.loaders.concat([
  { test: /\.js$/,
    exclude: /(node_modules|bower_components)/,
    loader: "react-hot-loader/webpack" }
]);

// ...

app.use(webpackDevMiddleware(compiler, {
  publicPath: config.output.publicPath,
  noInfo: true,
  stats: {
    colors: true
  }
}));

app.use(webpackHotMiddleware(compiler, {
  log: console.log,
  path: '/__webpack_hmr',
  heartbeat: 10 * 1000
}));
```

My component files are definitely being transformed, here is a snippet:

```js
// ...
module.exports = connect(
	  (state, props) => ({
	    breakpoints: _getBreakpoints(state)
	  }),
	  dispatch => bindActionCreators(actions, dispatch)
	)(Breakpoints);
	

	 ;(function () { /* react-hot-loader/webpack */ if (process.env.NODE_ENV !== 'production') { if (typeof __REACT_HOT_LOADER__ === 'undefined') { return; } if (typeof module.exports === 'function') { __REACT_HOT_LOADER__.register(module.exports, 'module.exports', "Breakpoints.js"); return; } for (var key in module.exports) { if (!Object.prototype.hasOwnProperty.call(module.exports, key)) { continue; } var namedExport = void 0; try { namedExport = module.exports[key]; } catch (err) { continue; } __REACT_HOT_LOADER__.register(namedExport, key, "Breakpoints.js"); } } })();
	/* WEBPACK VAR INJECTION */}.call(exports, __webpack_require__(17)))
```

Finally, how I render the root component:

```js
function renderToolbox() {
  ReactDOM.render(
    React.createElement(
      AppContainer,
[...]

PR #278, opened 2016-05-02
added test case for createClass and createFactory

nfcampos:
- test case for createClass is passing
- test case for createFactory is failing

for #276


nfcampos:
@gaearon should I merge these failing tests before we add the fix?

gaearon:
Yeah, let’s do this.

PR #280, opened 2016-05-02
(wip) added eslint

### LINKED ISSUE #274: Enable ESLint
I’d go with airbnb config. Semicolons ftw 😄 .
If somebody gets to it, might as well enable Travis (#247).
This should be done before 3.0.

Don’t forget the development happens on the `next` branch.

nfcampos:
@gaearon want to review the config?

I still need to fix most of the files to conform to the config

closes #274


.eslintrc
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "airbnb",
+  "parserOptions": {
+    "ecmaFeatures": {
+      "experimentalObjectRestSpread": true
+    },
+  },
+  "rules": {
+    "no-underscore-dangle": ["error", { "allow": ["__REACT_HOT_LOADER__"] }],
+    "jsx-quotes": ["error", "prefer-single"]
```
gaearon: I’m fine with using double quotes everywhere.

test/.eslintrc
```diff
@@ -0,0 +1,19 @@
...
+    },
+  },
+  rules: {
+    "react/prefer-stateless-function": 0,
+    "react/no-multi-comp": 0,
+    "react/prefer-es6-class": 0,
+    "react/prop-types": 0,
+    "no-shadow": 0,
+    "new-cap": 0,
+  },
```
gaearon: We might also want to disable it completely for Babel test fixtures.

gaearon:
Looks good. I left a few comments.
You can try using `eslint --fix .` to get some of these fixed.

.eslintrc
nfcampos: using double quotes for jsx and single quotes for everything else isn't weird?

test/.eslintrc
nfcampos: I have, that's why I created a `.eslintignore` file

nfcampos:
Yeah --fix helps at least with the semicolons, adding semicolons in hundreds of lines manually would have been terrible

nfcampos:
@gaearon you wrote some of the files without `const`, `import`, stuff like that, any reason to keep them like that, since we're building everything with babel? I need to either change them or change the eslint rules

.eslintrc
gaearon: It’s weird but apparently it’s what everyone else is doing these days 😄

test/.eslintrc
gaearon: Ah, missed that one.

gaearon:
If we use ES modules everywhere, we should add the commonjs compat plugin so that `react-hot-loader/webpack` and  `react-hot-loader/babel` keep working as they are.

nfcampos:
add the plugin to babel?

.eslintrc
nfcampos: double quotes it is then 😄

gaearon:
Yeah, this one: https://www.npmjs.com/package/babel-plugin-add-module-exports. Not sure if it works well though. Never tried it.

nfcampos:
hmm, not worth it then I think. basically I'm getting a bunch of `error  'use strict' is unnecessary inside of modules` from this rule http://eslint.org/docs/rules/strict + `airbnb`  config adding `parserOptions: {sourceType: 'module' }` to the eslint config.  So all of the non-es6 modules files in src/ are being interpreted as being modules and thus not needing 'use strict'. but i dont think it's possible to disable parserOptions on a single file. should i just disable the strict rule?

gaearon:
I’m a bit confused. I meant that we can use ES modules everywhere in the code. We just need to make sure that the CommonJS compat code is generated as part of Babel compilation for entry points like `/babel` and `/webpack`. Does this make sense? Or are you saying this makes ESLint fail?

nfcampos:
what i'm saying is that it currently works great the way it's written so instead of changing everything to es6 modules just to appease eslint it's easier to selectively disable the offending rule, no?

gaearon:
>what i'm saying is that it currently works great the way it's written so instead of changing everything to es6 modules just to appease eslint it's easier to selectively disable the offending rule, no?

Works for me 👍

nfcampos:
Alright, I've finished fixing all the linting errors. Give it a look if you want, I think I didn't break anything. Let's merge this soon to avoid merge conflicts

src/AppContainer.dev.js
```diff
@@ -54,38 +54,46 @@ class AppContainer extends Component {
...
       );
     }
+
+    return undefined;
   },
-  children: function (props) {
+  children(props) {
     if (React.Children.count(props.children) !== 1) {
-      return new Error(`Invalid prop "children" supplied to AppContainer. Expected a single React element with your app’s root component, e.g. <App />.`);
+      return new Error('Invalid prop "children" supplied to AppContainer.' +
```
gaearon: There’s a missing space after the period in the error message. The sentences will collide.
Also a style nit: let’s put a newline right after the opening paren.

src/AppContainer.dev.js
```diff
@@ -54,38 +54,46 @@ class AppContainer extends Component {
...
+        'Expected a single React element with your app’s root component, e.g. <App />.');
     }
-  }
+
+    return undefined;
```
gaearon: Lol this rule is so nitpicky. But I can live with it 😄

src/webpack/tagCommonJSExports.js
```diff
@@ -1,4 +1,6 @@
-;(function () {
+/* global __FILENAME__ */
```
gaearon: Kinda silly to leak these into the generated code. Can we add some naïve code to cut comments in the Webpack loader?

gaearon:
The error message needs to be fixed but the rest is up to your judgement.
:shipit:

src/AppContainer.dev.js
nfcampos: Good catch

src/AppContainer.dev.js
nfcampos: Yeah this one looks pretty silly in some places :)

src/webpack/tagCommonJSExports.js
nfcampos: I'll do that later

PR #285, opened 2016-05-04
Fixed typo this.props.prop -> this.props.props

leidegre:
From what I've been able to tell, this was not a bug in 3.0.0-alpha.11 but surfaced in 3.0.0-beta.1


nfcampos:
You're right, thanks!

PR #287, opened 2016-05-04
patch createFactory

### LINKED ISSUE #277: Doesn’t work with React.createFactory()
The issue is explained here: https://github.com/gaearon/react-hot-loader/issues/276#issuecomment-216302815. It seems like we’d need to monkeypatch `React.createFactory` because it bypasses `React.createElement` and uses the internal `ReactElement` module directly.

nfcampos:
fixes #277 

looks ok @gaearon ?


gaearon:
Great, thanks!

PR #293, opened 2016-05-11
Grammar fixes.

hedgerh:
@gadicc Made a few grammar fixes, mostly.  Opened a new branch so I didn't overwrite any unpushed work you may have.  Feel free to merge this into `gh-pages-next` after you've reviewed it!

### DIFF
--- a/getstarted.md
+++ b/getstarted.md
@@ -3,7 +3,7 @@ layout: page
 title: Get Started
 ---
 
-React Hot Loader is a plugin that allows instantaneous live refresh without losing state while editing React components.
+React Hot Loader is a plugin that allows instantaneous live refresh, without losing state, while editing React components.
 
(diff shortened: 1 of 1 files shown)


gadicc:
@hedgerh, all great!  Thanks for being so considerate :D  But I think next time just commit on the branch... let's both just commit whenever we finish up... if there ever is a conflict, we can manually fix it.  We can always just write a note in the PR if we're working on something big and don't want others to touch it.

hedgerh:
@gadicc Sounds good.  I figured as much, just wanted to bounce it past you first :D

gadicc:
Most considerate ever :)  Generally, I prefer what you did for my own projects, but I think for this, where we're both busy and we want to progress quickly, better to just commit.  I think I already took that liberty with some stuff you wrote :)  Better to keep things moving and revert a commit if problems do arise, especially since we're going to squash the commits at the end anyways.

PR #302, opened 2016-05-23
Change `NODE_ENV === production` to `module.hot`

### LINKED ISSUE #301: AppContainer complains because NODE_ENV === 'development'
I’ve just updated react-hot-loader to 3.0.0b2 in a project of mine (it’s gr8) and ran into something funky.

`AppContainer` was complaining about "react-hot-loader/patch" not running on a server rendered build of my project. Initially I thought I had to wrap `<App />` in `AppContainer` only if `module.hot` was set to true, but then I did some light digging and saw that `AppContainer` conditionally loads `AppContainer.dev.js` or `AppContainer.prod.js` based on the value of `NODE_ENV`.

I was doing some server rendering debugging and obviously didn’t need hot reloading, but I did need `NODE_ENV` to be set to development so I could see error messages.

All that to say, is there a better way to tell if hot reloading is desired that `NODE_ENV`? I’m a real dummy when it comes to the magical mystery of hot reloading, but can you check for `!module.hot` instead of `process.env.NODE_ENV === 'production'`?

If checking `NODE_ENV` is the best way to go, I think an addendum to the “react-hot-loader/patch didn’t run” message would be useful… something about `NODE_ENV` being set to development.

meyer:
Fixes #301, but I’m not sure if this is 👌 .

### DIFF
--- a/src/AppContainer.js
+++ b/src/AppContainer.js
@@ -2,8 +2,8 @@
 
 'use strict';
 
-if (module.hot) {
-  module.exports = require('./AppContainer.dev');
-} else {
+if (!module.hot && process.env.NODE_ENV === 'production') {
   module.exports = require('./AppContainer.prod');
+} else {
+  module.exports = require('./AppContainer.dev');
 }
--- a/src/AppContainer.js
+++ b/src/AppContainer.js
@@ -2,8 +2,8 @@
 
 'use strict';
 
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./AppContainer.prod');
-} else {
+if (module.hot) {
   module.exports = require('./AppContainer.dev');
+} else {
+  module.exports = require('./AppContainer.prod');
 }
--- a/src/babel/index.js
+++ b/src/babel/index.js
@@ -30,7 +30,7 @@ module.exports = function plugin(args) {
   const { types: t } = args;
 
   // No-op in production.
-  if (process.env.NODE_ENV === 'production') {
+  if (!module.hot) {
     return { visitor: {} };
   }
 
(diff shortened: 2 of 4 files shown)


src/babel/index.js
```diff
@@ -30,7 +30,7 @@ module.exports = function plugin(args) {
   const { types: t } = args;
 
   // No-op in production.
-  if (process.env.NODE_ENV === 'production') {
+  if (!module.hot) {
```
gaearon: This will not be true for the Babel plugin itself.

src/webpack/tagCommonJSExports.js
```diff
@@ -2,7 +2,7 @@
 
 ;(function register() { // eslint-disable-line no-extra-semi
   /* react-hot-loader/webpack */
-  if (process.env.NODE_ENV !== 'production') {
+  if (module.hot) {
```
gaearon: Similarly, `module.hot` will not be true inside Webpack loader itself, only in app's code.

src/AppContainer.js
```diff
@@ -2,8 +2,8 @@
 
 'use strict';
 
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./AppContainer.prod');
-} else {
+if (module.hot) {
```
gaearon: Let’s leave both checks here. `if (module.hot) && process.env.NODE_ENV === 'production'`

src/patch.js
```diff
@@ -2,8 +2,8 @@
 
 'use strict';
 
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./patch.prod');
-} else {
+if (module.hot) {
```
gaearon: Same here: let’s check both.

calesce:
👍  Looks good to me, thanks!

PR #308, opened 2016-06-19
Added missing generators to starterkits documentation

weblogixx:
Hi @gaearon,

here are some more projects using your (awesome!) loader. Hope they fit your requirements (the added generator is based on the react-webpack-template, so technically, it is a duplicate, but uses it just as a base for installation).


calesce:
Thanks!

Issue #312, opened 2016-06-27
error rendering failure covers up source error


rosskevin:
`react-hot-loader@3.0.0-beta.2`

I have a coding error in my application, yet it was not easily discovered because the `errorReporter` seems to be failing to render it. 

I receive these messages in the log:
```
Warning: React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components). Check the render method of `AppContainer`.

Uncaught Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object. Check the render method of `AppContainer`.
```

I edited my local [`AppContainer.dev.js`](https://github.com/gaearon/react-hot-loader/blob/next/src/AppContainer.dev.js#L52) to output the actual source error:

```javascript
console.log("error: ", error, " typeof: ", typeof error)
console.log("errorReporter: ", this.props.errorReporter, " typeof: " + typeof this.props.errorReporter)
return React.createElement(this.props.errorReporter, { error: error });
```

And I receive the source error and additional information:

```
error:  TypeError: Cannot read property 'replace' of undefined
    at module.exports (eval at <anonymous> (http://localhost:3808/webpack/application.js:8112:2), <anonymous>:14:19)
    at module.exports (eval at <anonymous> (http://localhost:3808/webpack/application.js:8106:2), <anonymous>:10:21)
    at new JwtToken (eval at <anonymous> (http://localhost:3808/webpack/application.js:8100:2), <anonymous>:47:43)
    at Security.getJwtToken (eval at <anonymous> (http://localhost:3808/webpack/application.js:8052:2), <anonymous>:196:27)
    at Security.ensureInitialWorkingTime (eval at <anonymous> (http://localhost:3808/webpack/application.js:8052:2), <anonymous>:123:12)
    at App.componentWillMount (eval at <anonymous> (http://localhost:3808/webpack/application.js:4764:2), <anonymous>:126:26)
    at App.componentWillMount (eval at <anonymous> (http://localhost:3808/webpack/application.js:1542:2), <anonymous>:44:30)
    at ReactCompositeComponentWrapper.performInitialMount (eval at <anonymous> (http://localhost:3808/webpack/application.js:4452:2), <anonymous>:336:12)
    at ReactCompositeComponentWrapper.mountComponent (eval at <anonymous> (http://localhost:3808/webpack/application.js:4452:2), <anonymous>:244:21)
[...]

PR #314, opened 2016-06-29
Fix RedBox require and add a test

nathanmarks:
This addresses #312 .

Right now in the console we're seeing an error crop up because `AppContainer.dev` is attempting to render the Redbox module export but the component sits on `.default` due to babelification and `AppContainer` is using `require()`.

### DIFF
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -2,7 +2,7 @@
 
 const React = require('react');
 const deepForceUpdate = require('react-deep-force-update');
-const Redbox = require('redbox-react');
+const Redbox = require('redbox-react').default;
 const { Component } = React;
 
 class AppContainer extends Component {
--- a/test/AppContainer/AppContainer.dev.js
+++ b/test/AppContainer/AppContainer.dev.js
@@ -1,7 +1,7 @@
 import './setup';
 import React, { Component } from 'react';
 import expect, { createSpy } from 'expect';
-import { mount } from 'enzyme';
+import { mount, shallow } from 'enzyme';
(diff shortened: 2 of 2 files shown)


EloB:
Please merge this because it's very hard to develop right now with the beta. The error output without this is also pointing you at the wrong code! So I have to remove the `<AppContainer>` from the render function when ever I get an error that I don't know where it's located.

peterbe:
My `index.js` now looks like this:
```javascript
const rootEl = document.getElementById('root');
// If you get errors like:
// Warning: React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components). Check the render method of `AppContainer`.
// then comment out these lines and uncomment the ones below.
ReactDOM.render(
  <AppContainer>
    <App />
  </AppContainer>,
  rootEl
);
// ReactDOM.render(
//   <App />,
//   rootEl
// );
```
:(

blesswinsamuel:
My temporary fix:

```jsx
const mount = document.getElementById('app');

render(
  <AppContainer errorReporter={({ error }) => {throw error}}>
    <App store={store} />
  </AppContainer>,
  mount
);
```

quaninte:
Please merge this one 👍

jiripudil:
Until this is merged, you can still use redbox-react for error reporting, just import it yourself correctly and overwrite the default `errorReporter` prop:

```js
import RedBox from 'redbox-react';
// or
const RedBox = require('redbox-react').default;

render(
    <AppContainer errorReporter={RedBox}>
        <App />
    </AppContainer>
);
```

Also, bump :)

Issue #242, opened 2016-04-18
Investigate what it would take to support function properties


gaearon:
e.g.

```js
class App extends Component {
  handleClick = () => {
    this.setState(...)
  }
}
```

One way to approach this would be to hoist the bodies of the function properties into hidden methods on the prototype with mangled names. This would be unobservable to the user but make them available for `react-proxy` to see and hot reload.

nfcampos:
> hoist the bodies of the function properties into hidden methods on the prototype with mangled names
can you do that without a babel plugin?

gaearon:
I don’t think so. But then again, you can’t have class properties without Babel anyway?

nfcampos:
yeah but then aren't we back to needing to identify classes that look like react components at compile time?

gaearon:
No, this transform is harmless for any class. I just mean turning

```js
class Stuff {
  doSomething = (e) => {
    this.x(e.stuff())
  }
}
```

into

```js
class Stuff {
  doSomething = (e) => {
    return this.___doSomethingqwertyu234567(e);
  },
  ___doSomethingqwertyu234567(e) {
    this.x(e.stuff())
  }
}
```

I’m not sure if there is anything in the spec that would make it a bad idea but it’s worth giving it a shot. The only special case I see so far is `arguments` object which should still end up being the one in `constructor`. To be honest we can bail completely when we see `arguments` used.

nfcampos:
Hmm, I see, we might hit some edge cases related to the differences between regular functions and arrow functions, so we should bail out on use of `super`, `new.target`, and `arguments` as you said, and we also need to convert arrow functions without curly braces to regular functions with body
```js
class Stuff {
  doSomething = (e) => this.x(e.stuff())
}
```

```js
class Stuff {
  doSomething = (e) => {
    return this.___doSomethingqwertyu234567(e);
  },
  ___doSomethingqwertyu234567(e) {
    return this.x(e.stuff())
  }
}
```
which means it might be better to translate it to
```js
class Stuff {
  doSomething = (e) => {
    return this.___doSomethingqwertyu234567(e);
  },
  ___doSomethingqwertyu234567(e) {
    return ((e) => this.x(e.stuff()))(e)
  }
}
```
this way the function would still be an arrow function... on the other hand we're creating a function on every invocation...

gaearon:
We need to create it once, otherwise it’s slow.

[...]

Issue #313, opened 2016-06-29
[3.0] Saving the value of 'this' is broken in class constructor but works with transform-es2015-classes


Kovensky:
The code in the following gist (which happens to be based on the example at  https://github.com/facebook/react/issues/7094#issuecomment-228931737) explains the issue in more detail:

https://gist.github.com/Kovensky/0ce87560e4b3c93cd77475fdcf94b610

Using `react-hot-loader@3.0.0-beta.2`. Note that the component in that gist is not inside an `AppContainer`, but the "real world" version of it ultimately is, and that one also has the same error behaviour.

The same problem also happens when using `transform-class-properties` and arrow function properties, as they're bound at construction time.

Kovensky:
This seems related to #242. If it's not supported at all, I wonder, then, why did it (appear to?) work when `transform-es2015-classes` was used...

jkeljo:
I'm seeing this as well. The React docs recommend this pattern of reassigning functions in the constructor for a couple of purposes:

- [binding handlers in ES6 classes](https://facebook.github.io/react/docs/reusable-components.html#no-autobinding)
- [using PureRenderMixin in ES6](https://facebook.github.io/react/docs/pure-render-mixin.html)

But even the presence of `react-hot-loader` is enough to break the pattern (you don't actually need to make a code change). From playing around in the debugger, it looks like react-hot-loader is cloning the objects in a way that doesn't update the binding of already-bound member functions.

I created a repro repo [here](https://github.com/jkeljo/react-hot-loader-and-es6-function-assignment-bug) before I realized that this was the same issue.

MarshallOfSound:
Came across this today while doing some work integrating `react-hot-loader` into a pure Electron dev workflow.  In Electron you only need to target latest stable Chromium which means that classes are already implemented natively.  This issue is much more prominent in that space because people just don't use the class transform as it isn't required.

If someone can point be in a good direction I'll be happy to dive into this at some point and get it sorted.

wkwiatek:
@MarshallOfSound it's the issue of [react-proxy](https://github.com/gaearon/react-proxy). If you wanted to start with this, just try to do a simple proxy object (with `react-proxy@next`), and play with a dubugger. We'll dive into `react-proxy` more once we have docs ready for RHL as it's the highest prio now.

MarshallOfSound:
[...]

Issue #263, opened 2016-04-28
Make sure we’re not breaking code by inserting a function


gaearon:
Relevant to: https://github.com/webpack/webpack/issues/2399
We probably should put `;` before and after the `(function () { ... })()`.
This might be necessary both for Babel and Webpack parts.

A pull request is welcome.

gaearon:
Fixed in `react-hot-loader@3.0.0-beta.0`.

PR #322, opened 2016-07-13
Add transform to support arrow function class properties

calesce:
I implemented this based off of the discussion in #242. I started with a [failing test case](https://github.com/calesce/react-hot-loader/blob/40e5e364b00a874014b09ab6e47a399b4e9e9a09/test/AppContainer/AppContainer.dev.js#L238) where class properties didn't get hot-reloaded.

I still need to add the case for arrow functions without bodies, but I figured I'd go ahead and open up the discussion.

Also, I added a `.babelrc` to the `AppContainer` test directory to include the Babel plugin, so I could test the class property case. While the existing test cases are still passing, I'm not sure whether running the transform will change how they work, since they rely on registering/tagging components manually.


calesce:
Ok, I fixed the case for arrow functions without block statement bodies. I also found a case where it failed, because the `value` field of a class property is nullable (which seems to happen when adding [`propType` annotations to class components via Flow](https://flowtype.org/docs/react.html#defining-components-as-reactcomponent-subclasses)).

gaearon:
Wow! Thanks for working on this.
I’ll definitely check it out soon.

ro-savage:
Awesome work @calesce !

We tried this out and found it broke when we had methods with default params. It seemed to work fine in all the other cases in our project.


Example:
```jsx
class defaultParamsComponent extends React.Component {

  defaultParam = (param = 'test') => {
    return param
  }

  render() {
    return (
      <div>
        {this.defaultParam()}
      </div>
    )
  }
}
```

Gives the error
> Property arguments[0] of CallExpression expected node to be of a type ["Expression","SpreadElement"] but instead got "AssignmentPattern"

@gaearon We'd love to see this implemented. Hot-reloading is pretty broken for us because we are using arrow function methods.

calesce:
Thanks for the feedback @ro-savage! I also just noticed it breaks when destructuring arguments, eg:
`onClick = ({ target }) => {}`. I should iron those two cases out tonight.

ro-savage:
@calesce I updated our react-hot-loader with your latest changes and its working perfectly for us. Thanks heaps!

calesce:
@ro-savage no problem!

I now realize that this type of transform won't work if a developer were to change the number of arguments when changing a class property function, since the argument count won't be reloaded. Which isn't good. I think something like this might work instead:

before:
```js
class App {
  foo = (bar, baz) => {
    return bar + baz
  }
}
```

after:
```js
class App {
  foo = (...params) => {
    return this.__foo__REACT_HOT_LOADER__(...params)
  }

  __foo__REACT_HOT_LOADER__(bar, baz) {
    return bar + baz
  }
}
```

That way, the number of arguments won't matter, so the developer can add/remove args and it should reload fine.

HriBB:
I tried your fork, and it works on a small app, but not on a larger application. I'm switching back to RHL1.3 for the moment. I hope you figure it out. Would love to help, but this is a bit over my head ATM.

calesce:
Hey thanks! Any details on where/how it doesn't work?

HriBB:
@calesce will test it again tomorrow and report. Basically all I need is class arrow functions to hot-reload. I use them heavily throughout my app and without them being hot-reloaded is pretty much useless for me. For now I switched to binding them in constructor, and now they do reload, so I don't have to reload the app anymore :)

ro-savage:
@HriBB, can you find out what is not hot reloading?

I am using it on an app that solely uses arrow functions and contains >200 components and haven't ran into any problems.

HriBB:
@calesce @ro-savage I must be doing something wrong, since even basic hot reload does not work. I'm not sure if I installed @calesce's fork properly. This is what I did.

```
git clone git@github.com:calesce/react-hot-loader.git
cd react-hot-loader
git checkout class-properties
npm install
cd ../my-project
npm link ../react-hot-loader
npm run client:dev
```

App loads fine, everything works except hot reload. I am using `webpack-dev-server` and `resolve.alias`. 

 When I change some file, I get the standard output in the console, but nothing happens. 
```
[HMR] Waiting for update signal from WDS...
[WDS] Hot Module Replacement enabled.
[WDS] App updated. Recompiling...
[WDS] App hot update...
[HMR] Checking for updates on the server...
[HMR] Updated modules:
[HMR]  - 1511
[HMR]  - 1510
...
[HMR]  - 1568
[HMR]  - 1666
[HMR] App is up to date.
```

But nothing is changed. It works as expected with npm version of `react-hot-loader@3.0.0-beta.2`

Any ideas? I'm looking at the RHL source, but I'm not yet familiar with it and so I have no idea how to debug this ...

This is my [config](https://gist.github.com/HriBB/489bbaa8160145309348d2f5cb75295f)

HriBB:
Ahh I finally got it to work. I guess that `npm link` is not the right way to install forks. Here's what I did. BTW I'm using node `v6.3.1`

```
git clone git@github.com:calesce/react-hot-loader.git
cd react-hot-loader
git checkout class-properties
npm install
rm -rf node_modules
npm install redbox-react react-proxy global react-deep-force-update
cd ..
cp -r react-hot-loader my-project/node_modules
cd my-project
npm run client:dev
```

@calesce @ro-savage how did you install the fork? I would really like to know so that I don't run into the same problems in the future.

It looks like it's working. I will do some more testing over the next few days and report any problems here ;)

calesce:
@HriBB: I think you needed to `npm run build` first.

The way I tested was copy-pasting the built files into my app's `node_modules/react-hot-loader` (where I already had beta3 installed).

`npm link` is probably the recommended/easier way, but I've only used it a few times so I'm not comfortable with it yet :wink:

HriBB:
@calesce yeah I know. It only worked if I manually copied files into my project's `node_modules/react-hot-loader` folder (after running `npm run build` in RHL3 fork).

If I used the `npm link` approach, I ended up with `node_modules/react-hot-loader/node_modules/[all RHL3 dependencies]` and hot reload would not work.

This is out of the scope of this PR, but does anyone know how to properly link forks?

carlosyasu91:
Any updates on this? I tried copying the built lib folder from the react-hot-loder repo on the `class-properties` branch into my project's `node_modules` but it didn't work, I get this error ` The following modules couldn't be hot updated: (They would need a full reload!)`, any ideas on what to try?

calesce:
@carlosyasu91 It's hard to say, any way you can push up a minimal project reproducing the issue?

test/AppContainer/AppContainer.dev.js
```diff
@@ -181,6 +181,211 @@ function runAllTests(useWeakMap) {
...
+            render() {
+              return <span onClick={this.handleClick}>new render</span>;
+            }
+          }
+          RHL.register(App, 'App', 'test.js');
+          wrapper.setProps({ children: <App /> });
+        }
+
+        wrapper.find('span').simulate('click');
+        expect(spy).toHaveBeenCalledWith('bar');
```
nfcampos: @calesce it might be worth it to test that the method is replaced without remounting the component, eg. like it is done here https://github.com/gaearon/react-hot-loader/blob/next/test/AppContainer/AppContainer.dev.js#L182

test/AppContainer/AppContainer.dev.js
```diff
@@ -181,6 +181,211 @@ function runAllTests(useWeakMap) {
...
+
+          render() {
+            return <span onClick={this.handleClick}>old render</span>;
+          }
+        }
+        RHL.register(App, 'App', 'test.js');
+
+        const wrapper = mount(<AppContainer><App /></AppContainer>);
+        wrapper.find('span').simulate('click');
+        expect(spy).toHaveBeenCalledWith('foo');
```
nfcampos: same as above about testing that it doesnt remount

package.json
```diff
@@ -50,8 +50,10 @@
   "devDependencies": {
     "babel-cli": "^6.7.5",
     "babel-core": "^6.7.6",
+    "babel-eslint": "^6.1.2",
     "babel-preset-es2015": "^6.6.0",
     "babel-preset-react": "^6.5.0",
+    "babel-preset-stage-1": "^6.5.0",
```
nfcampos: we could use stage-2 instead of stage-1 but it doesn't really matter

nfcampos:
@calesce this looks awesome! what do you think of adding more isolated tests that do not include the `es-2015` or `stage-1` babel presets, to see what out plugin is actually doing on its own. something like this https://gist.github.com/nfcampos/275ada99b630ee9524b374b0588b36c3

test/AppContainer/AppContainer.dev.js
calesce: Agreed, doesn't hurt.

package.json
calesce: Yeah I think they were still stage-1 when I first made the PR 😉

calesce:
@nfcampos thanks for the review!

I agree on not needing run the ES2015/stage1 transforms on the babel fixtures, the expected output is more readable that way.

nfcampos:
@calesce one other thing i just remembered, we might have to leave alone class properties that use `arguments` in the body of the function, because the way babel compiles them (see http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=stage-2&code=class%20Foo%20%7B%0A%20%20bar%20%3D%20(a%2C%20b)%20%3D%3E%20%7B%0A%20%20%20%20console.log(arguments)%0A%20%20%20%20return%20a(b)%3B%0A%20%20%7D%0A%7D ) they are the `arguments` of the `constructor`, and the way we compile them they are the arguments to the method we create on the class.

test/babel/fixtures/class-properties/block-body/expected.js
```diff
@@ -0,0 +1,21 @@
...
+
+class Foo {
+  bar = (...params) => _this.__bar__REACT_HOT_LOADER__(...params);
+
+  __bar__REACT_HOT_LOADER__(a, b) {
+    return a(b);
+  }
+
+}
+;
```
nfcampos: not related with this PR I think but what's up with this `;` and the other one at the end?

nfcampos:
@calesce take a look at this, I think it addresses what I mentioned above of opting out on use of `arguments` or `new.target` https://github.com/calesce/react-hot-loader/pull/1 sorry about the confusion of opening a PR against a PR, didn't know how else to do this, I think if you merge my PR this Pr will be updated with my changes

test/babel/fixtures/class-properties/block-body/expected.js
calesce: I assumed it was Babel weirdness, I was making the tests pass here. According to [AST explorer (highlight the semicolon)](https://astexplorer.net/#/wK0pmqT3vX) that's an additional empty statement, I'll see if I'm accidentally creating one.

calesce:
@nfcampos no problem, I think I would've have to add you to my fork's collaborators.

I'll take a look at why there are additional empty statements being created tonight, then this should be good to go 👍

calesce:
@nfcampos The semicolon insertion comes from #263, so not related.

Actually, it'd be nice if we could split out multiple plugins, so we can test them in isolation, but still be under a "root" `react-hot-loader/babel` plugin.

janv:
I have encountered an issue with this when using _static_ class properties

```
class Foo extends React.Component {
  static bar = () => 'baz'
}
```

Processed throught the react-hot transformation, `bar` will be removed from the class.

I have illustrated this in https://github.com/janv/react-hot-static-bug

calesce:
@janv: you're right, we're stripping the static keyword off of the properties. thanks for the repro!

janv:
I ran into another issue with this.
I recently tried removing the es2015 preset from our babelrc in development, since we're doing all our development in Chrome whcih should not need that preset.

We have Components nested two levels deep, that are forwarding callbacks which we create as arrow function class properties.

Something like

```js

class A {
  render(){
    return <B onClick={this.handleClick}/>
  }

  handleClick = () => this.setState(...)
}

class B {
  render(){
    return <button onClick={this.handleClick}/>
  }

  handleClick = () => this.props.onClick(...)
}
```

`setState` in A complains that the component is unmounted.

That should mean someone somewhere is keeping a reference to an original implementation of a function. I'll try to come up with a minimal executable example. Also, no sure why this only started happening after we removed babel-preset-es2015 yet

test/babel/fixtures/class-properties/block-body/expected.js
```diff
@@ -0,0 +1,21 @@
+var _this = this;
+
+class Foo {
+  bar = (...params) => _this.__bar__REACT_HOT_LOADER__(...params);
```
janv: Just noticed this line.
Wouldn't a call to `foo.bar()` be forwarded to `window.__bar__REACT_HOT_LOADER__` instead of the class instance? Since the first line binds `_this` to the global object and not the class instance?

janv:
I manages to write a minimal app that exhibits the problem:
https://github.com/janv/react-hot-static-bug/tree/arrow-bug

(Notice this is a **new branch** on the repo I used before to illustrate the issue with static class properties)

calesce:
Might be worth noting that we already were having problems when not using Babel to transpile classes (#313).

calesce:
@janv thanks again for the repro.

Yeah, that definitely seems like the same problem as #313. The issue remains even if you replace the class properties in your example with normal class methods.

I'm kind of surprised that "native" classes behave differently from how Babel transpiles them. The fix will probably have to come from [react-proxy](https://github.com/gaearon/react-proxy).

nfcampos:
@calesce yep I believe the fix would have to be in react-proxy, I played around with it on friday but haven't figured out what exactly is different about native `class` that makes this fail

calesce:
Weirdly, it looks like `react-proxy` [has test coverage](https://github.com/gaearon/react-proxy/blob/next/package.json#L10) for native classes.

nfcampos:
@calesce yeah I noticed, not sure why the tests are not failing but this is

PR #326, opened 2016-07-20
Fix hmr API Doc typo

stevenjob:
#257


calesce:
Also fixed by #349, but thank you!

PR #341, opened 2016-08-14
Update README.md

k3a:
added my favourite lightweight starter (react-webpack-babel) into which I also contributed


calesce:
Thanks!

Issue #337, opened 2016-08-07
[3.0.0-beta.2] RHL breaks when a component imports another component of the same filename


tanguylebarzic:
## Example

Using the webpack loader (without Babel), this breaks (with an unhelpful React error "Uncaught Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object. Check the render method of `AppContainer`."):

**In src/team/component.tsx**
`
import TeamList from '../list/component';
`

where _../list/component.tsx_ is another file.

This doesn't:

**In src/team/component.tsx**
`
import TeamList from '../list/componentbli';
`

where _../list/componentbli.tsx_ has exactly the same content as _../list/componentbli.tsx_. I'm using Typescript here, but I don't think it's related to the issue.

## Possible fix
My understanding is that we're overriding the registration of the first file by one of its imports.

I was able to fix it by changing the following line in react-hot-loader/lib/webpack/index.js:
```
const appendText = tagCommonJSExportsSource.replace(
    /__FILENAME__/g,
    JSON.stringify(path.basename(this.resourcePath))
  );
```

by:
```
const appendText = tagCommonJSExportsSource.replace(
    /__FILENAME__/g,
    JSON.stringify(this.resourcePath)
  );
```

This way, we register the two components under differents ids.

Another way:
```
const appendText = tagCommonJSExportsSource.replace(
    /__FILENAME__/g,
    JSON.stringify(path.relative(this._compiler.context, this.resourcePath))
  );
```
which makes for a nicer path but could have side effects (including accessing private webpack variables).

I don't know if the babel transform suffers from the same issue (it seems to use file.opts.filename, which may have the same issue, but I haven't tested it).

I can do a PR if you think this is a reasonable fix, tell me if there's a better way to fix it.
Thanks for this great lib!

pepjo:
I'm using babel and facing the same issue

aweber1:
I'm using webpack and seeing similar. In my case, I have a component with file name `AppMain.js` and a reducer elsewhere in the codebase with file name `AppMain.js`. These two files do not import or reference each other in any way, they just share the same name.

Thanks @tanguylebarzic for the fix!

calesce:
Fixed by #347.

calesce:
Out in `3.0.0-beta.4`!

PR #347, opened 2016-08-23
Fix the import of a component from a component with the same base name

tanguylebarzic:
Fix https://github.com/gaearon/react-hot-loader/issues/337, without accessing internal webpack variables.


calesce:
Cool, I played around with this, and it works in a few simple cases. I'd prefer to have this kind of stuff tested, but we don't have tests on the Webpack loader anyways (something to improve!).

Either way, since #337 was directly affecting at least a few people, and I don't see how it can break other use cases, I'll go ahead and merge it. Thanks!

PR #348, opened 2016-08-23
render() -> ReactDOM.render()

nahuel:


calesce:
Covered by #349, but thank you!

PR #349, opened 2016-08-23
Fixing the broken V3 example in the docs

jameskraus:
The V3 docs include a broken example with a hanging "/>" and the incorrect call to ReactDOM.render . The latter cost me an hour to notice. This is a small fix to those small problems.


calesce:
Thanks!

PR #372, opened 2016-09-20
fix async arrow functions

### LINKED ISSUE #370: react-hot-loader/babel creates invalid code with async arrow functions
On a class (that is _not_ a React component), there is a method defined through an arrow function:

```js
show = async (id, type) => {
  const { data } = await apiRequest(/*...*/)
```

Without the es2015 preset (only stage-2, react, and react-hot-loader/babel), this gets transformed to:

```js
_createClass(ClassName, [{
    key: '__show__REACT_HOT_LOADER__',
    value: function __show__REACT_HOT_LOADER__(id, type) {
      var { data } = await apiRequest(/*...*/);
```

This creates an `await` statement not inside an `async` function, which is invalid syntax.

This happens with `beta.4`, but rolling back to (the previously working?) `beta.3` does not seem to help.

jquense:
fixes #370


calesce:
This looks great!

Quick nitpick: can you move that async test to `test/babel/fixtures/class-properties/`?

nfcampos:
Yes this looks great! Do you mind adding a test for an async arrow function like this `async () => await 2` with no body?

calesce:
Made the changes we asked for, since I remembered that GitHub lets collaborators push to forks now. Thanks again @jquense!

jquense:
Ya thanks! sorry I didn't get a moment to make those changes myself. many thanks

PR #374, opened 2016-09-21
Prevent the Babel plugin from breaking the output of babel-node

### LINKED ISSUE #359: babel-node broken by 3.0.0-beta.3
The entry in the .bablrc plugins: "react-hot-loader/babel" is needed for the hot loading but it breaks the babel-node repl -  everything in the repl just returns undefined. Removing the plugin  restores the functionality making this a 100% reproducible issue for me.

calesce:
Closes #359.

The [tagger code](https://github.com/gaearon/react-hot-loader/blob/bc263de6d123f9f12b5eccfe3054feb948e7a3fd/src/babel/index.js#L7-L15) inserted by the babel plugin at the end of every file is an IIFE, which happens to break `babel-node` output when it's loaded. This happens because the code is added to every line that is evaluated by the REPL, so the lines now end with `(function() { return; })()`, which makes the REPL output `undefined` for every line.

By changing the IIFE from being wrapped in parens to a variable declaration, the output is no longer `undefined` for every line.

The tests pass and from manual testing it doesn't break anything. @nfcampos want to take a look?

Thanks to @loganfsmyth for the help on the Babel slack!

### DIFF
--- a/src/babel/index.js
+++ b/src/babel/index.js
@@ -4,6 +4,9 @@ const buildRegistration = template(
   '__REACT_HOT_LOADER__.register(ID, NAME, FILENAME);'
 );
 const buildSemi = template(';');
+
+// We're making the IIFE we insert at the end of the file an unused variable
+// because it otherwise breaks the output of the babel-node REPL (#359).
 const buildTagger = template(`
 var UNUSED = (function () {
   if (typeof __REACT_HOT_LOADER__ === 'undefined') {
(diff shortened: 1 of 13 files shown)


loganfsmyth:
Glad it worked out!

nfcampos:
@loganfsmyth this looks great! do we want to leave a comment in the code saying why we did this seemingly weird thing (assigning the result of a function that doesn't return anything to a variable we don't otherwise use)?

PR #381, opened 2016-09-26
more passing tests for class properties transform

nfcampos:
- test for class properties that are not functions (are skipped)
- test for class properties that are functions, but not arrow functions (are skipped)
- test for class property with same name as class method


calesce:
Cool, looks good.

PR #382, opened 2016-09-26
Opt out of transforming static class props

nfcampos:
tests for static class properties and change to opt out of tranforming them. 

I don't think there's any reason for us to transform those, they are already hot reloadable i believe


calesce:
Cool, looks good to me. Thanks @janv for pointing out the issue!

PR #393, opened 2016-10-07
add babel-plugin-transform-class-properties to class-properties babel fixtures

nfcampos:
- using only babel-plugin-syntax-class-properties produces invalid output, see https://github.com/gaearon/react-hot-loader/pull/322#discussion_r82390063


nfcampos:
@janv if you look at the output of fixtures of the babel tests on this branch you can see that `this` should be correct

janv:
👍 
Thanks a lot for this.
I also managed to create an example-app for my problem in #322

PR #404, opened 2016-10-21
Add TypeScript definitions

jupl:


calesce:
Hey, thanks for taking the time to do this. I think it'd be better to have type definitions in [definitely-typed](https://github.com/DefinitelyTyped/DefinitelyTyped), rather than being maintained by us, so I'd recommend making a PR there.

@nfcampos or @hedgerh, do you agree/disagree?

nfcampos:
I'm not familiar with Typescript enough to know whether the definitions belong here or in the `definitely-typed` repo.

Kovensky:
It looks like DefinitelyTyped [prefers](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/README.md#create-a-new-package) that libraries bundle their own types:

> If you are the library author, or can make a pull request to the library, bundle types instead of publishing to DefinitelyTyped.

There are definitions in DefinitelyTyped now (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/12632), though, so this could be closed.

calesce:
Thanks @Kovensky. Even though we have a small API, I still think it's better to have in `DefinitelyTyped`, and the typedefs are now there. Thanks again for your effort, @jupl.

PR #434, opened 2016-12-01
Installing a next version of react-hot-loader

bboysathish:
`npm install --save-dev react-hot-loader` installs 1.3.0 version so changing it to `npm install --save-dev react-hot-loader@next`, after merging to master need to revert back.


calesce:
Thanks! Can you just change this to be in the "A Big Update is Coming" section, perhaps at the end?

calesce:
Thanks!

PR #453, opened 2016-12-25
Update Troubleshooting.md

unkleho:
Just a small change I noticed when I was problem solving a npm link issue with react-hot-loader. https://github.com/gaearon/react-hot-boilerplate/blob/master/webpack.config.js#L27 (line 27) no longer exists, pretty sure it is https://github.com/gaearon/react-hot-boilerplate/blob/master/webpack.config.js#L22.
Cheers


calesce:
Thanks!

PR #460, opened 2017-01-17
TypeScript support for AppContainer

mutsys:
I was able to successfully implement the new react-hot-loader in a TypeScript/React project using the latest and greatest (React Hot Loader 3, Webpack 2.2, TypeScript 2.2, React 15.4.2, ReactRouter 3) and wanted to contribute back to the project. This commit contains a short TypeScript definition file with the necessary type information for `AppContainer`.

However, it is worth noting that the type definition for `AppContainer` is not sufficient on its own. While untyped JavaScript (ES5/6) is totally unconcerned if an undeclared object `module` is present, the TypeScript compiler chokes on it. Another ambient type definition must be provided to tell TypeScript about the existance of `module`, its property `hot` and its method/function `accept`. This is what works for me:

`webpackModule.d.ts`

```javascript

declare var module: WebpackModule;

interface WebpackModule {
    hot?: {
        accept: (dependency: string, callback: () => void) => void;
    }
}
```

It doesn't matter what you call this file, but it must exist in the same directory as the `.tsx` file that invokes `AppContainer`. Furthermore, this additional definition cannot live in the same file as the type declaration for `AppContainer` because it does not live in the `react-hot-loader` module.

In an ideal world, I guess this additional type definition would be provided somehow by Webpack itself, seeing as this is really a Webpack concern.  I poked around in `@types/webpack` but this is focused on webpack compiler configuration. Neither `@types/webpack-dev-middleware` or `@types/webpack-dev-server` seem to be a proper home either. On the other hand, React Hot Loader is the only project that I have seen so far that references `module.hot` outside of Webpack itself. 

So is it appropriate to discuss how React Hot Loader might somehow provide this additional type definition or would you recommend that I send a PR to the Webpack team to lobby for its inclusion there instead? Another alternative would be to send you a PR for `react-hot-boilerplate` that contains the type definition there.


aikoven:
[`@types/webpack-env`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/webpack-env/index.d.ts) is what you're looking for.

aikoven:
Also, consider adding this to `package.json`:
```js
"typings": "./index.d.ts",
```
This would allow TypeScript to discover typings automatically when you import `"react-hot-loader"`.

mutsys:
@aikoven `typings` property added to `package.json` now. Thank you for pointing me to `@types/webpack-env`, greatly appreciated. Still getting my feet wet in the type declaration labyrinth...

calesce:
See #404, we'd rather have TS stuff in DefinitelyTyped than in this repository.

PR #464, opened 2017-01-19
Fix #427

### LINKED ISSUE #427: Will this convention be supported?
I don't want to use `this.handleSliderMouseDown = this.handleSliderMouseDown.bind(this)` because this is too verbose. Also I don't want to apply the experimental syntax, `property-initializers`.

So, I've been binding function by declaring an arrow function inside of constructor.

```js
 constructor (props) {
    super(props)

    this.state = {
      listWidth: props.status.get('noteListWidth')
    }

    this.handleSliderMouseDown = e => {
      window.addEventListener('mouseup', this.handleSliderMouseUp)
      window.addEventListener('mousemove', this.handleSliderMouseMove)
      this.setState({
        isSliderActive: true
      })
    }
}
```

It works well as long as i use only HMR without React hot loader.
If I use HMR,  instance methods aren't  refreshed because constructor won't be fired again.

Can I use this syntax and wait the fix? Or should I stop using it?

Rokt33r:
I fixed #427 .

This PR must be reviewed. I'm also going to test this fork by my new project.


Rokt33r:
I confirmed it works.

rhl-fix

calesce:
Thanks @Rokt33r, looks great! I'll review more closely tomorrow. Apologies for the delay.

test/babel/fixtures/issue-427/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
calesce: can you call this directory something like `arrow-function-in-constructor` rather than `issue-427`? Also, we could probably use a few more test fixtures, see [what we have for class properties](https://github.com/gaearon/react-hot-loader/tree/next/test/babel/fixtures/class-properties).

I'd also like at least one `AppContainer` test, since those are more integration-level tests that verify that RHL works, not just the Babel plugin. Here's a [reference example](https://github.com/gaearon/react-hot-loader/blob/next/test/AppContainer/AppContainer.dev.js#L244)

Rokt33r: I'll do gladly. But, I think I can do after the 2nd week of the next month.(I have to finish my graduate thesis.) Could you wait about 3 weeks?

calesce: No problem! Take your time.

Rokt33r:
Done! Please review this again. @calesce

ThisIsRuddy:
Please merge :(

montogeek:
@calesce It is ok to merge it?

calesce:
Yep, sorry for the delay, I burned out hard maintaining this project :smile:

I only had issues with the file structure but the rest looked fine :+1:

ThisIsRuddy:
❤️

PR #467, opened 2017-01-26
fix: use relative file path instead of absolute

christophehurpeau:
In my case, react-hot-loader register my files with the absolute path. webpack-hot-middleware uses relative path, so hot reload don't work.

```js
  __REACT_HOT_LOADER__.register(init, 'init', '/Users/..../app/src/index.js');
```

This patch fix it: 

```js
  __REACT_HOT_LOADER__.register(init, 'init', './src/index.js');
```

Or is there another way to fix it ?

### DIFF
--- a/src/babel/index.js
+++ b/src/babel/index.js
@@ -124,7 +124,7 @@ module.exports = function plugin(args) {
           buildRegistration({
             ID: id,
             NAME: t.stringLiteral('default'),
-            FILENAME: t.stringLiteral(file.opts.filename),
+            FILENAME: t.stringLiteral('./' + file.opts.sourceFileName),
           })
         );
       },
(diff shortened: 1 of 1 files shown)


calesce:
I haven't had issues with `webpack-hot-middleware`, would you mind sharing a project reproducing the issue? Also, this breaks all of the Babel fixtures, so you'd need to update those.

christophehurpeau:
@calesce Thank you for your response ! I'll try to do a simple test example tomorrow

christophehurpeau:
@calesce https://github.com/christophehurpeau/react-hot-loader-test

calesce:
@christophehurpeau have you seen our [docs for RHL 3](https://github.com/gaearon/react-hot-loader/tree/next/docs)? You'll need to add a `module.hot.accept` call in your root module, as well as wrap your root React component in an `<AppContainer`>.

christophehurpeau:
@calesce yes indeed thanks for your answer !

PR #478, opened 2017-02-06
Add typescript note to README

danzel:
It was suggested we add a typescript note in #276.

At the moment all the examples in the readme include babel, which you don't use with typescript.
I've put this in the migration to 3.0 section straight after the babel config as I think it fits there. Can't think up enough to say to give typescript its own section.
Hopefully this helps anyone using typescript who runs in to the readme,

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,6 +28,16 @@
 }
 ```
 
+- When using **typescript**, babel is not required, so your config should look like ([demo](https://github.com/Glavin001/react-hot-ts)):
+
+```js
+{
+  test: /\.tsx?$/,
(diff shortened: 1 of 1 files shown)


DimitryDushkin:
It's very helpful addition!

docs/README.md
```diff
@@ -28,6 +28,16 @@
 }
 ```
 
+- When using **typescript**, babel is not required, so your config should look like ([demo](https://github.com/Glavin001/react-hot-ts)):
```
calesce: Can you move this section to the bottom with its own header? Also, please capitalize TypeScript/Babel.

danzel: Done.

calesce:
Looks great, thanks!

PR #486, opened 2017-02-20
Workaround reference error when using async arrow functions in code

wkwiatek:


calesce:
We might need to update the tests for now so we're not always failing.

PR #487, opened 2017-02-22
Improve docs installation steps

wkwiatek:


README.md
```diff
@@ -16,39 +16,33 @@ Some nice things about it:
...
 * configure Webpack to use React Hot Loader for JS or JSX files.
 
-These steps are covered by **[the Getting Started guide](http://gaearon.github.io/react-hot-loader/getstarted/)**.
+These steps are covered by **[the Migration to 3.0 guide](https://github.com/gaearon/react-hot-loader/tree/master/docs#migration-to-30)**.
 
 If you'd rather stay with **Browserify**, check out **[LiveReactload](https://github.com/milankinen/livereactload)** by Matti Lankinen.
 
+## Known limitations
+
+- React Router v3 is not fully supported. If you want to get most of React Hot Loader, consider switching to [React Router v4](https://reacttraining.com/react-router/). If you want to understand the reasoning, it's good to start in [React Router v4 FAQ](https://github.com/ReactTraining/react-router/blob/v4/README.md#v4-faq)
```
calesce: I'm not sure we should push people to RR4 until it's been fully released. Also, there are plenty of workarounds for RR3 that we can supply.

docs/README.md
```diff
@@ -28,7 +28,7 @@
 }
 ```
 
-- 'react-hot-loader/patch' should be placed at the top of the `entry` section in your Webpack config.  An error will occur if any code runs before `react-hot-loader/patch` has, so put it in the first position.
+- 'react-hot-loader/patch' should be placed at the top of the `entry` section in your Webpack config. An error will occur if any app code runs before `react-hot-loader/patch` has, so put it in the first position (note: if you use any polyfills then it should go right after them).
```
calesce: Maybe give a code example with `babel-polyfill` specifically?

README.md
wkwiatek: I'm still wondering what's the best option here... RHL and RR v3 are not the best pair and because of that RHL feels like 'works sometimes' and many of the issues are strictly related to RR. I just think we should communicate this somehow. 

On the other hand v4 works much better (and seems to be pretty solid) but it's still in beta. What about just adding some warning that v4 is still not a production version?

docs/README.md
wkwiatek: Definitely, will do!

calesce:
Looks great!

Issue #462, opened 2017-01-19
[3.0] Ability to display errors in console instead of redbox


icek:
Errors displayed in Redbox are useless for me especially when it comes to line numbers.

When sourcemaps are used, Chrome console can show exact error location (eg: `SomeComponent.js:12:13`) and you can jump directly into that file and chrome will highlight line causing error. Redbox on the other hand show line number in a bundled file (eg: `bundle.js:33052:13`) if you are patient you can search for this line...

Example of how useless it is:
![screen shot 2017-01-19 at 03 15 52](https://cloud.githubusercontent.com/assets/150243/22091172/d3b58292-ddf5-11e6-951a-10bcfcc1d4b8.png)

I digged a little and found that there is a property `errorReporter` on `AppContainer` so I tried to implement custom component for redirecting errors back to console, but I don't know how to make it work. Here is my initial approach:

``` jsx
import React from "react";

export default class NormalConsoleErrorInsteadOfRedbox extends React.Component {
    componentDidMount() {
        console.error(this.props.error);
    }

    componentDidUpdate () {
        console.error(this.props.error);
    }

    render() {
        return null; // <div/>
    }
}
```

used like this:
``` jsx
ReactDOM.render(
    <AppContainer errorReporter="NormalConsoleErrorInsteadOfRedbox">
        <Root/>
    </AppContainer>,
    rootEl
);
```

and now I get:
```
Warning: Unknown prop `error` on <NormalConsoleErrorInsteadOfRedbox> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in NormalConsoleErrorInsteadOfRedbox (created by AppContainer)
    in AppContainer
```

React 15.4.2, react-hot-loader 3.0.0-beta.6

icek:
Nevermind, here is quick workaround:

```js
delete AppContainer.prototype.unstable_handleError;

React.render(
```

and same error in console:

![screen shot 2017-01-19 at 03 37 36](https://cloud.githubusercontent.com/assets/150243/22091735/b75dc5ac-ddf8-11e6-869d-0ede0bc83479.png)

jtomaszewski:
Same here. Thanks for the workaround @icek , it works for me for now.

calesce:
I'm thinking we can add an `<AppContainer>` without a `unstable_handleError` method. Until we either get better (or faster) source maps as well as better error boundaries in React, the Redbox is far from perfect.

AlexKvazos:
[...]

Issue #164, opened 2015-07-20
Catch errors inside render


gaearon:
This is something I demoed in [my React Europe talk](https://www.youtube.com/watch?v=xsSnOQynTHs) but have not released yet.

Basically when you make a mistake inside `render`, React Hot Loader can catch the error and display a “red screen of death” inline instead of crashing your app:

<img width="491" alt="screen shot 2015-07-20 at 19 21 34" src="https://cloud.githubusercontent.com/assets/810438/8780882/950edda4-2f14-11e5-8bdc-8c9290e55b79.png">

There is a [PR to React Hot API](https://github.com/gaearon/react-hot-api/pull/17/files) which is exactly what I showed at the conf, but before turning this into an officially supported feature, I want to make it more extensible. I want to make the error renderer customizable because we want to support React Native too. Some projects may want to make custom renderers that display the error on top of `body` instead on inline, etc.

Ideally it would be opt in, and the API would be something like `react-hot-loader?reporter=npm-module-name` where `npm-module-name` is some package that exports a React (or React Native) `<Reporter>` component as a default export. This component will be used as `<Reporter error={e} />` by React Hot Loader, when specified.

Here is an example of such component by @KeywordBrain: https://github.com/KeywordBrain/redbox-react.

If somebody wants to take a stab at it, please do! (But let us know in this thread.)

If not, I'll get around to implementing this in August.

gaearon:
Edit: fixed the screenshot. :-) El Capitan screenshotting is broken..

davidpfahler:
@gaearon Just to be clear: The next step here is to implement the opt-in error report API, not any specific reporter, right?

gaearon:
Yes, the next step is to modify React Hot Loader to read (and resolve via Webpack's `require.resolve` I think) the reporter from the `query` config and pass it to React Hot API, then modify React Hot API based on my current PR so it uses that reporter, or, if reporter is not specified, behaves like it used to behave (just throw).

syranide:
Intuitively this seems separate from `react-hot-loader`, I could imagine wanting *just* this. Or a special version of this in prod where the errors are also submitted elsewhere.

gaearon:
@syranide Interesting. Yeah this *could* be separate. What API do you imagine?

syranide:
[...]

PR #494, opened 2017-02-28
Remove redbox

wkwiatek:
Closes https://github.com/gaearon/react-hot-loader/issues/462

### DIFF
--- a/docs/TipsAndTricks.md
+++ b/docs/TipsAndTricks.md
@@ -2,22 +2,19 @@
 
 **How to get an error in your console too:**
 
-The `Redbox` errors are great to clearly see rendering issues, and avoiding an uncaught error from breaking your app.  But there are some advantages to a thrown error in the console too, like filename resolution via sourcemaps, and click-to-open.  To get the best of both worlds, modify your app entry point as follows:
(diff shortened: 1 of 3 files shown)


calesce:
Looks good, let's also remove [this doc](https://github.com/gaearon/react-hot-loader/blob/9146001a9e8f076ad36669daa3c2bcdfb1631905/docs/TipsAndTricks.md). We still have #164 open in case we want to reconsider catching errors later on.

AlexKvazos:
I don't think removing redbox is the fix, more likely we have to improve sourcemaps so that redbox can show useful information.

wkwiatek:
@calesce Could you check the idea?

mcrawshaw:
@wkwiatek @calesce anything holding this back?

alexilyaev:
Guys, please merge this, or add an option to disable the RedBox (ideally an RedBox opt-in option).  
This is a long standing issue and it's annoying.

The way it is now, it forces us to add a hacky line when using `AppContainer`:
```
delete AppContainer.prototype.unstable_handleError;
```

or set a custom `errorReporter`, based on:  
https://github.com/gaearon/react-hot-loader/blob/master/docs/TipsAndTricks.md

Both of which should not be something a developer needs to think about when trying to setup hot-reloading, especially since it's a complicated process as it is.

neoziro:
Hello, I am a new maintainer on the project (@gaearon). I think we can remove Redbox since it is not the responsibility of this project. We could give the choice to the user, but the natural way of reporting error is the console. I would like to merge it, what do you think?

PR #498, opened 2017-03-05
docs(getstarted): fix typo

cyberhck:
document.getElementById instead of elementById
Are we sure it's elementById and not getElementById?


wkwiatek:
Of course it's not meant to be like this :) 

Thanks for giving PR!

PR #509, opened 2017-03-09
AppContainer handleError output on console

romulof:
handleError captures the error and render it, not allowing the error to be printed on console.
Rendering error is beautiful, but not functional. Console prints stacktrace with clickable links and translating lines to original source with sourcemaps.


wkwiatek:
Please see a discussion here #462 and this PR: #494

neoziro:
Duplicate of #494

PR #540, opened 2017-04-08
replace 0x2019 character with regular single quote

frankwallis:
The console error message was using a right-quote [’] character instead of the normal [']. In my situation (piping from webpack to ASP.NET) the quote was then getting automatically converted into a regular quote and this was causing the source file itself to be invalid.

I need to find out exactly why this was happening on my end, but to avoid similar situations in the future I would suggest changing it to a regular quote.


frankwallis:
I have resolved this issue in my code, so feel free to merge or close as you see fit, many thanks.

neoziro:
@frankwallis if you solved it, I prefer to not merge it.

PR #542, opened 2017-04-08
fixes React.PropTypes deprecation

### LINKED ISSUE #541: support React 15.5 deprecations
### Description

Package currently still uses `propTypes` from `react`. As per `react@15.5` this has been deprecated and an external [package](https://github.com/reactjs/prop-types) can be used instead.
### Expected behavior

Package functions as anticipated this is just an update that needs to be made to the code base to stay current with the `react` ecosystem.
### Actual behavior

Following error message is showing in the browser console:
screen shot 2017-04-08 at 12 04 32 pm

### Environment

React Hot Loader version:
```json
"react-hot-loader": "3.0.0-beta.6",
```

1. `node -v`: `7.7.2`
2. `npm -v`: `4.1.2`

Then, specify:

1. Operating system: `MacOS Sierra 10.12.4`
2. Browser and version: `Google Chrome Version 57.0.2987.133 (64-bit)`

Opening a PR will reference Issue there and PR here.

rockchalkwushock:
# What did I do to the code base

- [x] bumped `react` & `react-dom` to `v15.5.3`.
- [x] added `prop-types` as _devDep_.
- [x] used `prop-types` package in `./src/AppContainer.dev.js`.

Passed all tests & linting locally. Made a test repo for testing changes live. Ran from `react-hot-loader` directory locally with no issues. Repo can be found [here](https://github.com/rockchalkwushock/test-playground).

Any further changes or comments please feel free to reach out! This is my first time contributing 😟 

Fixes #541

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,9 +63,10 @@
     "expect": "^1.16.0",
     "jsdom": "^8.4.1",
     "mocha": "^2.4.5",
-    "react": "^15.0.2",
+    "proptypes": "0.14.4",
+    "react": "15.5.3",
     "react-addons-test-utils": "^15.0.2",
-    "react-dom": "^15.0.2",
+    "react-dom": "15.5.3",
     "recompose": "^0.17.0",
     "rimraf": "^2.5.2"
   },
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -2,6 +2,7 @@
 
 const React = require('react');
 const deepForceUpdate = require('react-deep-force-update');
+const propTypes = require('prop-types'); // eslint-disable-line
 const Redbox = require('redbox-react').default;
 const { Component } = React;
 


package.json
```diff
@@ -63,9 +63,10 @@
     "expect": "^1.16.0",
     "jsdom": "^8.4.1",
     "mocha": "^2.4.5",
-    "react": "^15.0.2",
+    "proptypes": "0.14.4",
```
simonkberg: This is not the correct package, should be [`prop-types`](https://www.npmjs.com/package/prop-types) (v15.5.4)

src/AppContainer.dev.js
```diff
@@ -2,6 +2,7 @@
 
 const React = require('react');
 const deepForceUpdate = require('react-deep-force-update');
+const propTypes = require('prop-types'); // eslint-disable-line
```
simonkberg: ESLint tried to tell you you were importing a non-existing package :)

simonkberg:
I just had another look, and this package isn't using `React.PropTypes`. The deprecation warning comes from [`react-redbox`](https://github.com/commissure/redbox-react) which is a dependency in this package.

rockchalkwushock:
@simonkberg thanks for catching the typo that led to the wrong package being installed. The packages is using `propTypes`; but only in `AppContainer.dev.js` as I saw. Maybe I am misunderstanding it's use in that file? Can you elaborate on what you mean by:

> this package isn't using `React.PropTypes`.

 Many thanks for the feedback; I'm new to doing this stuff 👍

simonkberg:
@rockchalkwushock no worries, it's great that you're contributing! I came here looking into where I got my own deprecation warnings from :)

So, let me try to explain: there are propTypes [defined here](https://github.com/gaearon/react-hot-loader/blob/master/src/AppContainer.dev.js#L59-L70) as a static property on the `AppContainer` component. However, no actual proptypes from the `react` package are used - it's just a function here doing a manual validation of the `children` prop.

What has been deprecated is accessing `PropTypes` from the `react` package, so this component actually doesn't contain any deprecated code, as it doesn't use `React.PropTypes`, but defines its own validation function.

In the stack trace of the deprecation warning, it looks like the error comes from this module, but they are actually from [react-redbox](https://github.com/commissure/redbox-react), which this package is using. So the problem could only be resolved there.

simonkberg:
Oh, btw, it seems like a new version of react-redbox fixing this has already been released: https://github.com/commissure/redbox-react/releases

rockchalkwushock:
@simonkberg Thanks for the explanation and information on the subject. I was unaware that you could use `propTypes` as a static property and create your own validation using it. Very cool! Thanks for following up with me! Closing.

PR #555, opened 2017-04-28
fix: apply the current scope when invoking a function (fixes #554)

### LINKED ISSUE #391: Invalid 'this' binding with async arrow function class properties
Async arrow functions use an invalid 'this' binding when they appear as class properties, as long as the `react-hot-loader/babel` plugin is present at all.

Arrow functions transformed by react-hot-loader also seem to not respect the `spec` setting given to the `transform-es2015-arrow-functions` plugin; however setting `spec` to `false` does not help avoid the invalid code either.

I've added some snippets of the problematic code, and also some comments annotating both the input and the output.

``` js
class Scroller { // not a React component
  /* ... */
  loadMore = async () => {
    const { data } = await api.getAsync(this.nextUrl, null)
    this.nextUrl = data.next_url
    this.appendData(data)
  }

  // written to verify if it only affected async
  loadMorePromise = () => {
    api.getAsync(this.nextUrl, null)
    .then(({ data }) => {
      this.nextUrl = data.next_url
      this.appendData(data)
    })
  }
}
```

Snippets of the generated code; in the constructor:

``` js
    this.loadMorePromise = function () {
      // _newArrowCheck and .bind(this) added by the spec: true option
      _newArrowCheck(this, _this);

      return this.__loadMorePromise__REACT_HOT_LOADER__.apply(this, arguments);
    }.bind(this);

    this.loadMore = function () {
      _newArrowCheck(this, _this);

      var _ref2 = _asyncToGenerator(regeneratorRuntime.mark(function _callee2() {
        var _args2 = _arguments;
        return regeneratorRuntime.wrap(function _callee2$(_context2) {
          while (1) {
            switch (_context2.prev = _context2.next) {
              case 0:
                _context2.next = 2;
                return _this.__loadMore__REACT_HOT_LOADER__.apply(_this, _args2);

              case 2:
                return _context2.abrupt('return', _context2.sent);

              case 3:
              case 'end':
                return _context2.stop();
            }
          }
        }, _callee2, _this);
      }));

      // wrong Function.name; this one is probably a babel bug
      return function NAME(_x2) {
        return _ref2.apply(this, arguments);
      };
    }.bind(this)();
```

In the `_createClass` helper's argument list:

``` js
  {
    key: '__loadMorePromise__REACT_HOT_LOADER__',
    value: function __loadMorePromise__REACT_HOT_LOADER__() {
      var _this2 = this;

      // why is this arrow function not using .bind and _newArrowCheck?
      api.getAsync(this.nextUrl, null).then(function (_ref6) {
        var data = _ref6.data;

        _this2.nextUrl = data.next_url;
        _this2.appendData(data);
      });
    }
  }, {
    key: '__loadMore__REACT_HOT_LOADER__',
    value: function () {
      var _ref7 = _asyncToGenerator(regeneratorRuntime.mark(function _callee5() {
        var _ref8, data;

        return regeneratorRuntime.wrap(function _callee5$(_context5) {
          while (1) {
            switch (_context5.prev = _context5.next) {
              case 0:
                _context5.next = 2;
                return api.getAsync(_this5.nextUrl, null);

              case 2:
                _ref8 = _context5.sent;
                data = _ref8.data;

                // _this5 was not defined anywhere, runtime crash
                _this5.nextUrl = data.next_url;
                _this5.appendData(data);

              case 6:
              case 'end':
                return _context5.stop();
            }
          }
        }, _callee5, this);
      }));

      function __loadMore__REACT_HOT_LOADER__() {
        return _ref7.apply(this, arguments);
      }

      return __loadMore__REACT_HOT_LOADER__;
    }()
  }
```

### LINKED ISSUE #554: does not respect public field binding
### Description

What you are reporting:

* A bug

### Expected behavior

What you think should happen:

`handleClose` should be called with the current context of the component.

### Actual behavior

What actually happens:

`handleClose` is called with the old context of the container.

### Environment

React Hot Loader version:

^3.0.0-beta.6

1. `node -v`: v7.9.0
2. `npm -v`: v4.2.0

Then, specify:

1. Operating system: OS
2. Browser and version: Chrome v58

### Reproducible Demo

```bash
git clone git@github.com:gajus/sanity-test.git
cd ./sanity-test
npm run install
npm run dev

```

Click "Click me" button.

Expected result: `movie` prop is not null.
Curent result `movie` prop is null.

### Additional information

The lack of context binding is visible in the code transpilation difference without/ with react-hot-loader/babel https://gist.github.com/gajus/2d0867f326d5c69702f057c7cd7cf15f/revisions

gajus:
* fixes #554
* maybe fixes #391

### DIFF
--- a/src/babel/index.js
+++ b/src/babel/index.js
@@ -19,8 +19,14 @@ var UNUSED = (function () {
 
 const buildNewClassProperty = (t, classPropertyName, newMethodName, isAsync) => {
   let returnExpression = t.callExpression(
-    t.memberExpression(t.thisExpression(), newMethodName),
-    [t.spreadElement(t.identifier('params'))]
+    t.memberExpression(
+      t.memberExpression(t.thisExpression(), newMethodName),
+      t.identifier('apply')
+    ),
+    [
+      t.thisExpression(),
+      t.identifier('params'),
+    ]
   );
 
   if (isAsync) {
--- a/test/babel/fixtures/class-properties/block-body/expected.js
+++ b/test/babel/fixtures/class-properties/block-body/expected.js
@@ -1,6 +1,6 @@
 class Foo {
   constructor() {
-    this.bar = (...params) => this.__bar__REACT_HOT_LOADER__(...params);
+    this.bar = (...params) => this.__bar__REACT_HOT_LOADER__.apply(this, params);
   }
 
   __bar__REACT_HOT_LOADER__(a, b) {
--- a/test/babel/fixtures/class-properties/default-params/expected.js
+++ b/test/babel/fixtures/class-properties/default-params/expected.js
@@ -1,6 +1,6 @@
 class Foo {
   constructor() {
-    this.bar = (...params) => this.__bar__REACT_HOT_LOADER__(...params);
+    this.bar = (...params) => this.__bar__REACT_HOT_LOADER__.apply(this, params);
   }
 
   __bar__REACT_HOT_LOADER__(a = "foo") {
--- a/test/babel/fixtures/class-properties/destructured-params/expected.js
+++ b/test/babel/fixtures/class-properties/destructured-params/expected.js
@@ -1,6 +1,6 @@
 class Foo {
   constructor() {
-    this.bar = (...params) => this.__bar__REACT_HOT_LOADER__(...params);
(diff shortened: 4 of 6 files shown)


gajus:
This doesn't fix the issue.

PR #558, opened 2017-05-02
allow output error message on browser console panel

AepKill:
```javascript
<AppContainer handleError={false}>....</AppContainer> // not use errorReporter handle error , all error message will output on browser console panel
<AppContainer handleError={true}>....</AppContainer> // use errorReporter handle error
```
handleError default value is true.


neoziro:
@AepKill not the way we choose to do it see #494

PR #563, opened 2017-05-09
Fixed test execution on Windows

dizel3d:
Comparison tests were failed due to line endings: `babel` produces JS-files with LF, but `git` checkouts all text files with OS default line endings (CRLF for Windows).

Now `git` is forced to checkout JS-files with LF.


neoziro:
@dizel3d thanks

PR #567, opened 2017-05-17
Updated Mocha

TheLordHighExecutioner:
npm installs were throwing a lot of warnings for deprecated dependencies. Updating mocha fixed it easily.


neoziro:
@TheLordHighExecutioner thanks but we will update all dependencies, not just mocha.

PR #572, opened 2017-05-30
Adding another possible error message to the Troubleshooting.md

andrerpena:
I was debugging this error:
   
    (function register() { /* react-hot-loader/webpack */ if (process.env.NODE_ENV !== 'production') { if (...

The above was running on my browser. `process` was, of course, undefined, so `env` couldn't be read.

The given solution for the message `Uncaught TypeError: Cannot read property 'NODE_ENV' of undefined`, described in this document, was also the solution for my message, which was `Uncaught TypeError: Cannot read property 'env' of undefined` so I decided to contribute to this document.


montogeek:
Thanks for the contribution @andrerpena 💃 !

PR #576, opened 2017-06-05
Update readme about create-react-app migration

zeakd:
As create-react-app released with webpack 2, there are little changes.


cdaringe:
still doing full refreshes on me with these changes, but i am using webpack@latest (v3.x)

neoziro:
We will take care of updating all docs at the same time. But thanks for PR.

PR #582, opened 2017-06-10
Update to latest redbox-react version

dcosson:
There's a noisy console error due to deprecated React.PropTypes in redbox-react. They have fixed it in later versions so react-hot-loader should use the latest version.

From #541


neoziro:
@dcosson we are not sure to keep it and we will update all dependencies at the same time. But thanks.

PR #588, opened 2017-06-15
add an express webpack 2 boilerplate example

zhenyulin:
currently it is still under //TODO on the [guidance page](https://gaearon.github.io/react-hot-loader/getstarted/), hope this will be useful for people who are still in the process of upgrading


neoziro:
@zhenyulin you are not owner or collaborator so it's not the good section to add it.

PR #594, opened 2017-06-21
added ts-react-boilerplate

sotnikov-link:
This boilerplate with tests on karma + jasmine + coverage and with sourcemaps for all configs. Boilerplate uses react-hot-loader.


neoziro:
@sotnikov-link thanks

PR #608, opened 2017-07-15
To catch a remount: Add debug information

theKashey:
The main goal of Hot Reloading - is to apply code changes on a live application.
But, If you can NOT preserve application state, including internal component state and component life cycle - it is better NOT to use Hot Loading features at all.

Unfortunately, It is not easy to make this feature work for any complex application - any @decorator, functional composition chain or too complex HoC - and everything will break.

This PR will solve this issue. It will add debug information, so a developer will understand which component is not patched and is causing local remount.

There are 3 types of components:
1. Registered ones. Work as they always do. Exists in `idsByType`.
2. Ones from node_modules. Work as they always do. Exists in `hasCreatedElementsByType`
3. Unregistered hot-reloader. Will be added by `signature` to `knownSignatures` for the first time, and reported as an issue for the second. But only after HMR event.

screen shot 2017-07-15 at 11 09 17 pm

HRM event will be caught by re-registration of any normal component. In this case some `different` components with the same signature (styled-components for example) will not soil console during startup.


theKashey:
Just improve a bit debug information. Now one can just click on function signature to found the `wrong` component.
Also I just wondering - should I suppress error messages in case Component does not have any children. Sometimes they are just annoying.
Pro - remount of component will not cause nested tree remount, as long there is no tree.
Con - remount exists and can case parasite side effects.

neoziro:
@theKashey very interesting, I like the idea. I keep it opened, I have to take a deeper look before merging it.

theKashey:
Hooray!

PR #619, opened 2017-08-09
Consistency

Toosick:
More consistent with https://github.com/gaearon/react-hot-loader/tree/master/docs#migration-to-30


neoziro:
@Toosick thanks

PR #628, opened 2017-08-31
Add two example start kits done in TypeScript and ES2015

codeBelt:
Add two start kits with the same code. One done in TypeScript and the other in ES2015 to show TypeScript is not that different or scary.

Example code on how to use: TypeScript, Universal (SSR), React Hot Loader 3, React Router 4, Redux, Redux Saga, Redux Form, Async Component Code Splitting, Hapi, Webpack 3.


neoziro:
@codeBelt thanks

PR #629, opened 2017-09-01
feat(AppContainer): support componentDidCatch for React v16

AndersDJohnson:
Add support for React v16 `componentDidCatch`, and proxy to that from React v15 `unstable_handleError`.


neoziro:
Do we need test for that? I think we can merge it.

PR #636, opened 2017-09-16
Fix #548

### LINKED ISSUE #548: Typescript Config Module: ES6 Causes State Reset
### Description

I'm not sure if this is specifically a react-hot-loader issue or some strange compatibility issue with typescript, but using Typescript to compile with `tsconfig.json`
```
{
  "compilerOptions": {
    "module": "ES6",
  }
}
```
will cause the entire component to re-create, destroying the state.

### Expected behavior

The state should remain

### Actual behavior

The entire component to re-create, destroying the state.

### Environment

Typescript version: 2.2.0

React Hot Loader version: 3.0.0-beta.6

1. `node -v`: v7.5.0
2. `npm -v`: 4.1.2

1. Operating system: macOS Sierra
2. Browser and version: Chrome  Version 57.0.2987.133 (64-bit)

### Reproducible Demo

https://github.com/jdolle/react-hot-loader-typescript-issue

asvetliakov:
When compiling with ES2015 modules, webpack stores them into ```__webpack_exports__``` variable instead of ```module.exports``` which will be undefined, thus registration won't occur. 

Tested with TS/Webpack3 and both commonjs/es2015 module output.

I'm not sure what this is correct way to fix

Fixes #548

### DIFF
--- a/src/webpack/tagCommonJSExports.js
+++ b/src/webpack/tagCommonJSExports.js
@@ -7,19 +7,25 @@
       return;
     }
 
-    if (typeof module.exports === 'function') {
-      __REACT_HOT_LOADER__.register(module.exports, 'module.exports', __FILENAME__);
+    /* eslint-disable camelcase, no-undef */
+    const webpackExports = typeof __webpack_exports__ !== 'undefined'
+      ? __webpack_exports__
+      : module.exports;
+    /* eslint-enable camelcase, no-undef */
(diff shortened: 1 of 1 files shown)


neoziro:
@asvetliakov yeah, I think Webpack support is broken since it uses "__webpack_exports__". I think this is the correct way to fix it.

PR #638, opened 2017-09-17
Upgrade dependencies and project

neoziro:
Hello, I am a new collaborator on the project. I use React for three years now, I am the author of [Shipit](https://github.com/shipitjs/shipit) and I give JavaScript trainings in France ([Smooth Code](https://www.smooth-code.com/).

My goal is to maintain this project and release a stable v3.

For my first PR, I did several things:

- Replace mocha by Jest
- Use snapshots for Babel tests
- Use babel-preset-env instead of deprecated babel-preset-es2015
- Upgrade project dependencies (especially source-map)
- Upgrade ESLint and use Prettier
- Target node 6 and node 8 on Travis

@gaearon you look like to be the main contributor on the project, I would be happy to have your quick review before merging it.

Aside, I merged all relevant PR on the project. Next steps will be:

- Take care of all opened issues
- Upgrade documentation
- Release a stable v3


src/AppContainer.js
```diff
@@ -1,9 +1,7 @@
 /* eslint-disable global-require */
 
-'use strict';
```
gaearon: I think we need `use strict` here. If ESLint says we don't, it's wrong.

src/AppContainer.prod.js
```diff
@@ -1,18 +1,17 @@
 /* eslint-disable react/prop-types */
 
-'use strict';
```
gaearon: Same. It's not an ES6 module. It needs `use strict`.

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "singleQuote": true,
+  "trailingComma": "all",
+  "semi": false
```
gaearon: Please let's keep semis 😛

neoziro: @gaearon OK 😤

gaearon: If you really insist we can remove. You're the maintainer now 😄

gaearon: It's the fate of all my projects now. I pass them on and people remove the semicolons...

src/AppContainer.js
neoziro: We transpile it using Babel, so why do we need to add it?

```sh
$ babel src/AppContainer.js                                                                                                                       
'use strict';

/* eslint-disable global-require */

if (!module.hot || process.env.NODE_ENV === 'production') {
  module.exports = require('./AppContainer.prod');
} else {
  module.exports = require('./AppContainer.dev');
}
```

gaearon: Oh okay.

.prettierrc
neoziro: It is not a real problem since Prettier does the job for us.

gaearon: Yea I know. Just funny.

PR #652, opened 2017-10-04
Remove proptypes from generated code.

AaronFriel:
PropTypes were deprecated in React 15.x and removed in 16.0. If there is still need for proptypes, the [prop-types](https://www.npmjs.com/package/prop-types) package should be used.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,9 @@
 {
   "presets": ["env", "react"],
-  "plugins": ["transform-class-properties"],
+  "plugins": [
+    "transform-class-properties",
+    "transform-react-remove-prop-types"
+  ],
   "env": {
     "test": {
       "plugins": ["transform-class-properties", "./lib/babel"]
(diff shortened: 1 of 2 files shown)


neoziro:
It is not necessary to remove PropTypes because they will not be included in production. If you take a look to .prod.js files there is no trace of `require('prop-types')`.

AaronFriel:
> It is not necessary to remove PropTypes because they will not be included in production. 

Well that's neat, but what about the people running `react-hot-loader` in development? The current situation is that it will throw an uncaught exception and redbox or stop rendering because the `PropTypes` export was removed in React 16: `React.PropTypes` is undefined. Merely defining PropTypes with `React.PropTypes` methods now results in errors. I've submitted an alternate patch to continue using PropTypes with the `prop-types` package, but they were deprecated.

neoziro:
OK I understand, it is relative to React 16. I was waiting for enzyme to be able to test it on React 16. Will do it soon.

AaronFriel:
I don't understand, you closed #653 and closed this issue saying, quote:

> Let's focus on #652.

Is this project dead?

PR #653, opened 2017-10-06
Update to prop-types package.

AaronFriel:
An alternative to gaearon/react-hot-loader#652, because right now React-Hot-Loader *does not work* with React v16, because `React.PropTypes` is undefined.


AaronFriel:
I can certify that this build does work, and it is the tests that are out of date and depend on deprecated React features. That's what this and #652 are about.

neoziro:
Let's focus on #652.

PR #655, opened 2017-10-07
Document HOC Troubleshooting

theKashey:
Leaving few notes in limitations and breaking troubleshooting into pieces.


neoziro:
Thank you!

PR #658, opened 2017-10-07
Stable v3 & React 16 compat

neoziro:
This PR is a first step to be compatible with React 16, some tests are not passing. I tried to investigate but I need help.


neoziro:
I thought it was related to enzyme and update issue but I tried with pure `ReactDom.render` and I have the same issues. Looks like [react-deep-force-update](https://github.com/gaearon/react-deep-force-update/) or [react-proxy](https://github.com/gaearon/react-proxy/) is not compatible with React 16. @gaearon do you have an idea?

gaearon:
Is it certainly using the very latest version of `react-deep-force-update`?

neoziro:
@gaearon yes 😕

neoziro:
@gaearon I dove into react-deep-force-update and it looks like stateNode is not updated.

https://github.com/gaearon/react-deep-force-update/blob/89506f42004843bd7297d44fdbd535fa2c9735cc/src/index.js#L83-L86

I think node of type "HostComponent" are not updated. How to force an update of these nodes?

neoziro:
@gaearon problem occurs only with stateless functional components, they are not updated and the function is not called. If I switch it by a class based component, it works.

neoziro:
@gaearon also I used the `react-deep-force-update` proposed in https://github.com/gaearon/react-deep-force-update/pull/9. I updated code, if you want to investigate, just switch to my PR and run `jest --watch test/HotContainer.dev.test.js`. If we make this test pass, it is good.

neoziro:
@gaearon I finally found the problem, in the following case, App will not be reloaded because element instance is not recreated. It was the case in React 15. I think it is a minor regression but it has to be noticed.

```js
const App = () => <div />
const element = <App />
ReactDOM.render(<HotContainer>{element}</HotContainer>, main)
```

ntucker:
Thanks for working vigilantly on React 16 compatibility @neoziro ! :)

neoziro:
@ntucker I try 🙂

PR #663, opened 2017-10-11
Fix incorrect AppContainer usage in docs

andrewcoelho:
The docs currently use this approach:
```
if (module.hot) {
  module.hot.accept('./containers/App', () => { render(App) });
}
```
But without showing an actual `render` function. This update fixes that discrepancy.


neoziro:
@andrewcoelho thank you!

Issue #378, opened 2016-09-21
Composing HoCs causes component state to reset


byronmwong:
First of all, thanks for the [great writeup](https://medium.com/@dan_abramov/hot-reloading-in-react-1140438583bf) on how hot reloading works. It allowed me to narrow down the issue I was seeing.

Here's a branch with a repro case: https://github.com/byronmwong/react-hot-boilerplate/pull/1. The counter will reset if you change anything in that file.

When multiple HoCs are used without intermediate assignments, the babel plugin isn't able to get a hold of the intermediate components that are getting created by the functions. I imagine this issue would get hit most commonly through use of `compose` for HoCs, e.g.

```
export default compose(
  withBackgroundColor('green')
  withPadding('10px')
)(MyComponent)
```

I'm not sure if this is solvable through the current static analysis approach, but I figured it was worth asking. Thanks!

nfcampos:
yeah I'm aware of this issue, but still haven't thought of a good way of detecting this in static analysis

wmertens:
I encountered this issue with a single decorator as well, the `@graphql` [decorator from react-apollo](https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L166-L202).

By assigning the decoration separately, hot-reload-with-state works.

I tried debugging by looking at the [react timeline](https://facebook.github.io/react/blog/2016/11/16/react-v15.4.0.html#profiling-components-with-chrome-timeline) but I got no farther than that the decorator render function is called and afterwards the decorated component is unmounted and remounted, maybe because the decorated component class reference changed?

calesce:
@wmertens yep supporting decorated components (and HoCs that are able to be detected statically) should be feasible, see discussion in #279.

rogersp:
I'm experiencing the same state reset behavior, when calling react-redux connect() on the component returned inside my higher order component function ([this line](https://github.com/rogersp/rhb-redux-hoc/blob/next/src/HOC.js#L23)).

Interestingly, the state does not reset when using the master branch of react-hot-boilerplate.

Repro examples:
- State IS preserved (base: react-hot-boilerplate/master): https://github.com/rogersp/rhb-redux-hoc/tree/master
- State is NOT preserved (base:react-hot-boilerplate/next): https://github.com/rogersp/rhb-redux-hoc/tree/next

To clarify, the state is maintained for anything that comes from the redux store, but any state the wrapped component tracks internally is reset on hot reload.

rogersp:
[...]

PR #665, opened 2017-10-11
 Use console.warn in warnAboutUnnacceptedClass

adi1133:
I think this error is related to #378, this is a non-fatal error (hot reloading is working), I prefer this to be a warning instead of an error.


neoziro:
You are right, a warning is sufficient! Thanks!

PR #669, opened 2017-10-12
Add support for disabling warnings for unaccepted components

### LINKED ISSUE #666: Component not accepted by Hot Loader if using react-refetch
Hi! congrats on v3!
I have been trying to switch from `react-transform-hmr` to `react-hot-loader@3`and facing some issues. It looks like I can't get Hot Loader to properly works with components wrapped with https://github.com/heroku/react-refetch HOC (and since my parent App component is also wrapped, none on the children components are accepted)

```
React Hot Loader: this component is not accepted by Hot Loader. 
Please check is it extracted as a top level class, a function or a variable. 
Click below to reveal the source location: 
 ƒ RefetchConnect(props, context) {
        _classCallCheck(this, RefetchConnect);

        var _this2 = _possibleConstructorReturn(this, _Component.call(this, props, context));

        _this2.version …
```

It looks really related to react-refetch (my App component is wrapped with multiple HOCs using `compose` from `recompose` and hot reload properly works as soon as I remove the `react-refetch` HOC). Any idea?

### LINKED ISSUE #667: Make warnings optional
Add a prop `warnings` to `AppContainer` that will disable warnings if set to `false`:

```js
<AppContainer warnings={false}>
  ...
</AppContainer>
```

Internally, this warnings prop will set `__REACT_HOT_LOADER__.warnings` to `false` in [AppContainer constructor](https://github.com/gaearon/react-hot-loader/blob/master/src/AppContainer.dev.js#L14).

In [patch.dev](https://github.com/gaearon/react-hot-loader/blob/master/src/patch.dev.js#L90) we can read this variable and display a warning or not.

Will solve #666

simonkberg:
Implemented according to instructions in #667. Wasn't sure how to properly re-create the issue in test, but I borrowed the test for the warning and made sure it was being silenced when `__REACT_HOT_LOADER__.warnings` is `false`.

The test for `AppContainer` simply checks that the global variable is being set.

Fixes #667
Fixes #666

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,16 @@ render((
 
 You'll also need to `npm install --save-dev redbox-react`.
 
+## Disable warnings
+
+React Hot Loader will by default emit a warning for components not accepted by the Hot Loader. If you want to disable these warnings, you can pass a `warnings` prop with the value `false` to `AppContainer`.
+
+```js
+<AppContainer warnings={false}>
+  ...
+</AppContainer>
+```  
+
 ## Starter Kit
 
 Provided by collaborators:
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -7,6 +7,14 @@ const { Component } = React
 class AppContainer extends Component {
   constructor(props) {
     super(props)
+
+    if (
+      props.warnings === false &&
+      typeof __REACT_HOT_LOADER__ !== 'undefined'
+    ) {
+      __REACT_HOT_LOADER__.warnings = false
+    }
+
     this.state = { error: null }
   }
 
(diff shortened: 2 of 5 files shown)


neoziro:
Very good! Thank you for implementing it!

Can you set RHL.warnings to true by default? For now it's undefined and it can be confusing.

simonkberg:
Thanks for reviewing. Pushed an update where I set `warnings: true` on default props and pass it on to RHL.warning. Let me know if this is what you had in mind, or if I should rather keep AppContainer as it was and put the default on `hooks` in `patch.dev` instead

neoziro:
Sorry, I was thinking about the default on `hooks` in `patch.dev`. I wasn't enough clear, sorry, can you remove it from `AppContainer` and just do it in `hooks` ?

simonkberg:
No problem at all! Adjusted accordingly

HosseinAgha:
@neoziro Could make a release? the warning is very annoying. Thanks.

PR #673, opened 2017-10-15
fix: revert arrow function in constructor support

neoziro:


theKashey:
Should we leave a memo about `problems` using Hot-loader with untranspiled ES6?

src/babel/index.js
```diff
@@ -23,37 +21,6 @@ const buildNewClassProperty = (
...
 }
 
-const buildNewAssignmentExpression = (
-  t,
-  classPropertyName,
-  newMethodName,
-  isAsync,
-) => {
-  let returnExpression = t.callExpression(
-    t.memberExpression(t.thisExpression(), newMethodName),
```
theKashey: So - you have an arrow function somewhere in class.
```class {
  test = () => { body }
}```
And babel transpile it as
```
 contructor() {
   this.test = () => {body};
 }
```
And then you transpile it as 
```
class{
   constructor(){
     this.test = (...params) => this.test_REACT_HOT_LOADER(...params);
   }
   test_REACT_HOT_LOADER(){
     body
   }
}
```

And here is the problem - in terms of ReactHotLoader `this` in constructor is always incorrect.

Propose - transpile it  
```
     this.test = (...params) => this.test_REACT_HOT_LOADER.call(REACT_HOT_LOADER_GET_THIS(this), ...params);
```
Where
```js
REACT_HOT_LOADER_GET_THIS = (that) => that && that.REACT_HOT_LOADER_INST ? that.REACT_HOT_LOADER_INST : that
```

And text react-proxy(or RHL proxy) can provide and keep up to date reference to the `current` instance from the first one via that secret field.

Or we can patch react-proxy to patch the original instances with the new ones, not to `proxy` the calls.

neoziro: I agree, but without a Lerna project containing all three projects (react-hot-loader, react-proxy, react-deep-force-update) this is not easy.

And yes we should have a memo for this kind of issues.

drcmda:
Would it be possible to release a quick revert to where beta.7 was?

neoziro:
@drcmda I will try to do it this week-end

PR #679, opened 2017-10-20
Customize scope for arrow functions

theKashey:
Fixes https://github.com/gaearon/react-hot-loader/issues/391 and (already closed) https://github.com/gaearon/react-hot-loader/issues/242

@neoziro, this is the same idea, that I mentioned in your PR.

Problem – react-proxy instances `ES5 classes` with correct `this`, but cannot repeat the trick with ES6 classes, as result new Components will be created with their own contexts.
In other word - `this` in arrow function will target the `real` component, but one can use only the proxied one.

Solution - store pointer to the proxy context inside instance, and use it.

This PR does include just a change of babel plugin, to work normally it requires changes in react-proxy, but version 3, used by RHL is not public :(

The one, who wants repeat the trick - just copy-paste code to the createProxyClass.js
```js
function linkInstance(instance, context){
  try {
    instance.__REACT_PROXY_CONTEXT = context.__REACT_PROXY_CONTEXT = context;
  } catch (err) {};
  return instance;
}

function proxyClass(InitialComponent) {
  // Prevent double wrapping.
  // Given a proxy class, return the existing proxy managing it.
  var existingProxy = findProxy(InitialComponent);
  if (existingProxy) {
    return existingProxy;
  }

  let CurrentComponent;
  let ProxyComponent;
  let savedDescriptors = {};

  function instantiate(factory, context, params) {
    const component = factory();

    try {
      return linkInstance(component.apply(context, params), context); // <----
    } catch (err) {
      // Native ES6 class instantiation
      const instance = new component(...params);

      Object.keys(instance).forEach(key => {
        if (RESERVED_STATICS.indexOf(key) > -1) {
          return;
        }
        context[key] = instance[key];
      });

      linkInstance(instance, context);  /// <-- not returning anything
    }
  }
```

Tested both in es5 and es6 environments – all green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,9 +43,10 @@
   "dependencies": {
     "global": "^4.3.0",
     "react-deep-force-update": "^2.1.1",
-    "react-proxy": "^3.0.0-alpha.0",
+    "react-stand-in": "^1.0.0-alpha-2",
     "redbox-react": "^1.3.6",
-    "source-map": "^0.6.1"
+    "source-map": "^0.6.1",
+    "stacktrace-js": "2.0.0"
   },
   "devDependencies": {
     "babel-cli": "^6.7.5",
--- a/package.json
+++ b/package.json
@@ -43,10 +43,9 @@
   "dependencies": {
     "global": "^4.3.0",
     "react-deep-force-update": "^2.1.1",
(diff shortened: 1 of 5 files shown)


theKashey:
Tests are dead without correct proxy.

PR #681, opened 2017-10-20
Add note about setting output.publicPath property

selbekk:
This PR adds a note about the importance of having the
`output.publicPath` property set to the correct value. Without it,
hot reloading will not work for nested routes like `/some/path`.

I'm up for a discussion about the wording or placement of this note - but I strongly feel that this should be mentioned in the docs. It has cost me a few hours of debugging through the last couple of months (I always forget how to fix this error when it shows up), and it would have been fairly simple to fix if somebody just told me it had to be set :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ module.exports = {
 }
 ```
 
+> Note: Make sure to set the `output.publicPath` property to `"/"` as well. Otherwise hot reloading won't work as expected for nested routes.
+
 4. Wrap your application into `<AppContainer>`, all children of `<AppContainer>` will be reloaded when a change occurs:
 
 ```js


neoziro:
@selbekk thanks, I think it is good like this.

adc17:
@selbekk does this mean `a.` `react-hot-loader` only works if you serve your static files from `/`? Or does it mean `b.` you have to declare the location of your static files at `output.publicPath`? 

The first time I read the readme, it sounded like you meant a, but now I've read your comment, it sounds more like you mean b.

selbekk:
@adc17 - I see what you’re saying. I definitely meant B - do you have any suggestions as to how we can change the wording? I’d be more than happy to create a new pull request! :-)

theKashey:
Why hot reloading will not work?

selbekk:
@theKashey it won't work because the hot reloader tries to fetch the reload from the relative path instead of the absolute path, where they are made available.

adc17:
@selbekk Thanks for clarifying, that's good news! 

I think your points are "don't forget to set it", and "don't forget the trailing slash". 

Facebook's CRA has this: `Webpack uses publicPath to determine where the app is being served from. It requires a trailing slash, or the file assets will get an incorrect path.`

Perhaps something along those lines to explain how it works, plus "so remember to set `output.publicPath` with a trailing slash in your webpack config".

theKashey:
@selbekk , @adc17 - not it's clear, thanks.
In my case it was a bit more tricky - there was no static public path. So, you can always set it on a client side:
```js
// find `your` script among all others on the page.
__webpack_public_path__ = myScriptPath; 
```

Issue #596, opened 2017-06-27
RHL is checking correctly but not updating


kimjuny:
Hot update checking seems to work correctly but not updating as expected(the changes I've made to Component not updated). Has this something to do with electron?

Min-repo [HERE](https://github.com/kimjuny/rhl-repo/tree/master).

[![enter image description here][1]][1]

<b>webpack.config.js</b>

```javascript
var path = require('path');
var webpack = require('webpack');
var HtmlWebpackPlugin = require('html-webpack-plugin');
    
module.exports = {
  entry: [
    'react-hot-loader/patch',
    path.resolve(__dirname, './src/index.jsx'),
  ],
  output: {
    publicPath: '/',
    path: path.resolve(__dirname, 'dist'),
    filename: 'bundle.js',
  },
  module: {
    rules: [
      {
        test: /\.js*/,
        include: [ path.resolve(__dirname, 'src') ],
        loader: 'babel-loader',
      }, {
        test: /\.less$/,
        use: [
          { loader: 'style-loader' },
          { loader: 'css-loader' },
          { loader: 'less-loader' },
        ],
      },
    ],
  },
  resolve: {
    alias: {
      'COMPONENTS': path.resolve(__dirname, './src/components'),
      'CONTAINERS': path.resolve(__dirname, './src/containers'),
      'MODELS': path.resolve(__dirname, './src/models'),
      'ROUTES': path.resolve(__dirname, './src/routes'),
      'SERVICES': path.resolve(__dirname, './src/services'),
      'THEMES': path.resolve(__dirname, './src/themes'),
    },
    extensions: ['.js', '.jsx', '.less'],
  },
  plugins: [
    new webpack.HotModuleReplacementPlugin(),
    
    new webpack.NoEmitOnErrorsPlugin(),
    
    new webpack.NamedModulesPlugin(),
    
    new HtmlWebpackPlugin({
      template: path.resolve(__dirname, 'src', 'index.html'),
    }),
  ],
  devtool: 'source-map',
  target: 'electron-renderer',
  devServer: {
    port: 8000,
    hot: true,
    historyApiFallback: true,
  },
};
```

<b>index.jsx</b>

```javascript
import React from 'react';
import ReactDOM from 'react-dom';
import { AppContainer } from 'react-hot-loader';
import Routes from './routes';
    
const main = document.createElement('div');
main.id = 'main';
document.body.appendChild(main);
    
const doRender = (Component) => {
  ReactDOM.render(
[...]

PR #700, opened 2017-11-22
Added { modules: false } step to the Getting Started docs

mildrenben:
I also had this issue: https://github.com/gaearon/react-hot-loader/issues/596

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,18 +23,19 @@ npm install --save react-hot-loader
 
 ## Getting started
 
-1. Add `react-hot-loader/babel` to your `.babelrc`:
+1. Add `react-hot-loader/babel` and set `{ modules: false }` in your `.babelrc`:
 
 ```js
 // .babelrc
 {
+  "presets": [["env", { "modules": false }],
   "plugins": ["react-hot-loader/babel"]
 }
 ```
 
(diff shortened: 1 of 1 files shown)


theKashey:
We can not add it yet:
1. RHL will not work with env preset.
2. "modules" affects only one thing - will webpack auto re-require "accepted" module, or you will have to call require by your own.

Actually, currently RHL's readme is not asking to enable harmony modules, but listed examples does not contain re-requiring updated code - and that is a mistake we should fix.

mildrenben:
RHL _does_ work with env though. I'm using it right now. It only works with `modules: false`.

theKashey:
@mildrenben if you will specify the target as a modern browser - RHL will blow up whole application due to some issues with non-transpiled arrow functions.
Anyway - you named the second step as the third one.
Please check numeration, and I'll be quite keen if you will include few words about different behaviours for different `modules` mode.

coffenbacher:
There's a typo here, missing `]` in the first line

```
{
  "presets": [["env", { "modules": false }],
  "plugins": ["react-hot-loader/babel"]
}
```

neoziro:
This isn't required in v4 anymore. https://github.com/gaearon/react-hot-loader/tree/next

PR #705, opened 2017-12-03
refactor: modernize packages

neoziro:
Use lerna, jest, and modern stuff.


theKashey:
Prettier is a magic.
Looking good, but I am not sure it will compile.
Anyway - good to merge, and I'll fix pipeline during my daylight time.

neoziro:
@theKashey I simplified tests, run prettier and added scripts to compile packages globally. I will find a way to run test on different Babel configurations.

theKashey:
Sure. Are you done for today?

neoziro:
@theKashey yes I will fix build and it is OK for me.

theKashey:
Ok. Ready to merge?

neoziro:
@theKashey no

theKashey:
Ok. Then it is my time to continue fixes.

neoziro:
@theKashey it should be OK now

neoziro:
@theKashey can you go on Google Hangouts please?

neoziro:
@theKashey ready to merge, diff is because Babel configuration is not fixed to a specify set of plugin `"node": "current"`. Nothing important.

neoziro:
@theKashey using a Babel "env" config standard we have some tests that are not passing. I think it is a problem because it is the most used preset.

theKashey:
What about keeping `boilerplate` inside? 
RHL is a tricky thing, and it is better to test change against some implementation, with proper instrumentation (babel or webpack).
In the same time - it is a good way to ship a boilerplate as an example.

Issue #650, opened 2017-10-03
HotLoader won't work after adding >=2 levels of HoC.


quangbuule:
Way to reproduce the issue:
- Using this example (`next` branch) https://github.com/gaearon/react-hot-boilerplate/tree/next
- Add `src/HoC.js`
```js
import React, { Component } from 'react';

export function provideHoC() {
  return WrappedComponent => {
    class HoC extends Component {
      render() {
        return <WrappedComponent {...this.props} />
      }
    }

    HoC.displayName = (WrappedComponent.displayName || WrappedComponent.name) + 'WithHOC';
    return HoC;
  };
}
```
- Modify `src/Counter.js`
```diff
 import React, { Component } from 'react';
+import { provideHoC } from './HoC';

-export default class Counter extends Component {
+class Counter extends Component {
   constructor(props) {
    super(props);
    this.state = { counter: 0 };
  }

  componentDidMount() {
    this.interval = setInterval(this.tick.bind(this), 1000);
  }

  tick() {
    this.setState({ counter: this.state.counter + 1 });
  }

  componentWillUnmount() {
    clearInterval(this.interval);
  }

  render() {
    return <h2>Counter: {this.state.counter}</h2>;
  }
}
+
+export default provideHoC()(provideHoC()(Counter));
```

- `npm start` and go to `http://localhost:3000`.
- Modify some text inside `render` method of `Counter`.
- The counter state will be gone.

neoziro:
We have to add this to documentation.

theKashey:
@neoziro. Never too late. I'll do it.

neoziro:
@theKashey done in #655

quangbuule:
Hard to fix on huge projects with tons of nested HoCs (since it is proper way to do HoCs). I created babel plugin to solve this:
https://www.npmjs.com/package/extract-hoc

theKashey:
@quangbuule - looking quite good, but, please, add tests and a bit more description.

neoziro:
This is good, we could eventually integrate it in the project if it is enough stable.

quangbuule:
I added some tests here: https://github.com/quangbuule/extract-hoc/blob/master/test/__snapshots__/babel.test.js.snap

mateoKaradza:
Still running into this issue, even after using extra-hoc babel plugin and manually extracting HoCs into separate components

Package.json: https://github.com/blazing-edge-labs/admin-playground/blob/master/package.json
[...]

Issue #703, opened 2017-11-29
State lost on code update


theKashey:
The major part of RHL is to understand how one can replace the old code, by the new. Currently, we have a quite approach:
 - in case of babel - `register` all top-level variables.
 - in case of webpack - all exported variable.
Each time React will try to update any component, not from this list - it will re-render tree, and you will get a state lost.
So HOC, decorators, functional compositions - you are drunk, go away. For many-many people, this is just blocker, and we have to solve it.

PS: State loss is not a problem. Mount/Unmount is a problem.

I've spiked possible solutions from many directions and got a one, which may work.
Move react tree traversal from react-deep-force-update to react-hot-loader, and perform an update by comparing the old and the new tries. It does support onUpdate/shouldUpdate now, but does not expose position inside a render tree.
One old tree and the new tree _should_ be the same, if some node got another set of children - you have to purge the old one in any case. But for RHL cases trees will be same, and in case of more `global` update HRM will update the whole page.

So the algorithm is simple:
1. Update the code.
2. before update traverse the old tree.
3. Call render method of a first Component. Fetch part of an old tree and try to find similarities.
 3.1 In case of named-components match - easy. It is ok - update proxies and go deeper.
 3.2 In case of named-components mismatch - easy. You can go deeper with state loss.
 3.3 In other cases try compare source code of components. If it is same - it is mostly safe to update proxies. Look like even having 100500 components with the same source code, but with the different scope variables, it will work fine.
4. ....
5. Profit.

This will fill the gaps between registered components and `fix` RHL :)

Look like it could work. @neoziro, I saw some crazy tree-traversal stuff in react-loadable, so could you leave few comments?

NeekSandhu:
Extremely relevant #714

neoziro:
Actually these are the ideas implemented in [React Hot Loader v4](https://github.com/gaearon/react-hot-loader/tree/next/). Thanks @theKashey!

PR #711, opened 2017-12-06
The Reconciler

theKashey:
This implements idea from #703 
Fixes: #650 (HoC), #666 (HoC), and any other HoC or decorator issues.

Idea - instead of relying on registering the variables - relay on react-tree.

Algorithm is simple
 - use something like react-deep-force-update to hydrate the render tree
 - `render` the tree back, swapping the old components, and the new ones on the fly.

Results
The good one - ANY component could be Hot-Reloaded. You can use complex HOC or decorators and still have RHL working🏆 
The bad one - all components should be wrapped by proxy. This actually does not affect performance at all, as long stand-in operates on pure js prototypical inheritance.

To activate reconciler - add `reconciler` prop to AppContainer.

PS: Not to be merged yet.


packages/react-hot-loader/src/patch.dev.js
```diff
@@ -1,77 +1,12 @@
...
-}
-
-let proxiesByID
-let didWarnAboutID
-let hasCreatedElementsByType
-let idsByType
-let firstInstances
-let knownSignatures
-let didUpdateProxy
+import {
```
theKashey: patch.dev was completely rewritten. Things like error reporting about failed registrations or failed replacement were removed, as long they are obsolete with reconciler.

neoziro:
Very good work! The idea is great! I think you have resolved the major issue of this project 😄.

packages/react-hot-loader/src/reconciler/ComponentMap.js
```diff
@@ -0,0 +1,61 @@
+export default class ComponentMap {
+  constructor(useWeakMap) {
+    if (useWeakMap) {
+      this.wm = new WeakMap()
```
neoziro: We could rely on WeakMap since we are targetting only modern browsers.

packages/react-hot-loader/src/AppContainer.dev.js
```diff
@@ -2,16 +2,20 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import deepForceUpdate from 'react-deep-force-update'
 import { getGeneration } from './updateCounter'
+import hydrate from './reconciler/traverse'
```
neoziro: hydrate / traverse

packages/react-hot-loader/src/reconciler/deepForceTraverse.js
```diff
@@ -0,0 +1,141 @@
...
+      }
+      return {
+        ...b[index],
+        type: child.type,
+      }
+    }),
+  }
+}
+
+const deepForceTraverse = (instance, stack) => {
```
neoziro: The goal of this function is to traverse a tree (stack) and swap all components is it right? I think the naming is not good.

packages/react-hot-loader/src/reconciler/traverse.js
```diff
@@ -0,0 +1,63 @@
...
+    do {
+      const childStack = {}
+      traverseTree(child, childStack)
+      stack.children.push(childStack)
+      child = child.sibling
+    } while (child)
+  }
+}
+
+export default function hydrate(instance) {
```
neoziro: The goal is to traverse React tree and build an AST of it in order to traverse it and swap all swappable components, yes?

theKashey:
Ideally this should be moved into react-mount, as a user defined class comparison function. And, as result we could even remove proxy, as long there will be no need to fake types. 
A single line change in react internals can make things must simpler.
How to sell this idea to the React team?

neoziro:
@gaearon can help us!

neoziro:
@theKashey we can't remove proxy because we have to keep state and variables. So even if we have a custom comparison function proxies are still needed.

gaearon:
Please avoid relying on React internal APIs if you can. We don't want to introduce more dependencies than already exist in `deep-force-update`.

I intend to work on officially supporting hot reloading in React in the beginning of next year. So it's best to coordinate our efforts.

theKashey:
@gaearon - does this mean, that we should not proceed with this PR, as long it uses some React internals, or we could proceed, as long it uses the same internals as deep-force-update?

gaearon:
If it uses the same exact internals it’s okay.

theKashey:
All `react internals` were move into a single (and quite small) file. Also refactored logic a bit (just perform dead code elimination).

theKashey:
@neoziro - just added a couple of tests. Coverage is not great, but enough for a start. 
It is much more important to test this solution against some real edge cases, I am not quite aware of.

Waive if you are keen to merge this to the lerna branch.

neoziro:
@theKashey Thanks! I could take a look on tuesday, not before.

gaearon:
I want to note that I still don't really understand what this is doing, so I'd appreciate a more in-depth explanation of why this works, why it is better than the existing approach, and what are the potential downsides and things that could break.

That said I don't want to block you on my understanding :-) If you're sure this is solid and give it a lot of testing in large codebases (I'm sure plenty of people would be happy to volunteer trying an alpha) then let's roll with it.

theKashey:
Ok, let me explain.

The biggest constraint of V3 is a literal unusability for lots of people. All these folks who use decorators, recompose, and quite happy about it - they are all off the boat.

V1 injects update code on the each-file-basic, thus allows it to perform point updates, and thus allows it to work with a complex code. But actually, it does not respect any complete code dependencies - module self-accept is a dangerous thing.

V3 performs the work much more gently, rendering from top to bottom, relying on `registered` classes to perform updates. This is the power, this is a weakness. It is just impossible to `register` everything, but you have to.

`Reconciler` is just another way to detect changes. 
The idea is simply - you have a rendered React Three, and you are updating a _single_ class. And you want only a single component to be updated (as V1 do).

Solution:
1. (hydrate)Traverse the existing tree, to store component type positions. Like - I know that Button was inside Form, inside Page12. This part actually borrows code from react-deep-force-update.
2. (hotReplacementRender)Traverse the new tree, to spot the changes to update proxies. I this part I have to explain in deep.

Reconsiler:
1. To make reconciler works RHL have to wrap all components with proxy. In real - not all, but all stateless, and all classes with injected "REGENERATE_METHOD". As long standin-proxy did not affect performance - we can do it.
2. Call render method on `real` component instance, obtained from hydrate.
3. Traverse the result, comparing component types.
3.1 In case of DOM node perform monkey patching (mergeInject) combining type from rendered child, and children from hydrated one.
3.2 In case of function compare the new and the old one
3.2.1 If types are equal - continue render
3.2.1 If types are equal from proxy point of view (`registed`) - continue render
3.2.1 If types are `swappable` - update proxy and continue render.

And `swappable` is the key of this solution.
In v3 `swappable` means that types have been exported from a same file and same name.
Here - it does mean - types were found in the same tree position, have the same display name and also some `similarity` in code.

`similarity` - full-text similarity, or 20% by Levenshtein distance (fuzzyCompare, up to O(n^2) ). For classes text similarity are tested on method level, allowing you add or remove one method, and change another. (has no effect on the `registered` types).

And there is a small trick here - reconsiler does not respect a possible change in props. It does not `render` the true react tree. It tries to re-render application as it was, to spot and swap changes.
If you are:
 - rendering a different component – it will don't enter that leaf
 - rendering component in a different order - it will swap types, and render them in the old order
 - rendering with different props, resulting a different nested tree - it will update as many proxies as it could, and the nested tree will be completely updated in any case.

So, this is not a real render, this is just another way to rehydrate hot updated code. 
It is still important to have a registered types, to perform major updates upon them, but now RHL can stand any decorator or function composition between these registered types.
Hooray?

packages/react-hot-loader/src/AppContainer.dev.js
```diff
@@ -99,6 +112,7 @@ AppContainer.propTypes = {
   },
   errorReporter: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
   warnings: PropTypes.bool,
+  reconciler: PropTypes.bool,
```
neoziro: I think we should always activate reconciler. Why making it optional?

packages/react-hot-loader/src/AppContainer.dev.js
```diff
@@ -99,6 +112,7 @@ AppContainer.propTypes = {
   },
   errorReporter: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
   warnings: PropTypes.bool,
```
neoziro: So we can drop this option isn't it?

theKashey:
RE:
```js
// from
compose(one, two, three)(Component)
// to
compose(one, three)(Component)
```
Reconciler will do it's best, and update all the proxies. But react-dom will not stand a tree change.
WE could invent a way, to rehydrate the old instances into a new components, but I'll prefer not to hack React (more deeply).

Issue #707, opened 2017-12-04
Next version: v4


neoziro:
This issue is designed to centralize discussions about the next version of React Hot Loader.

### Documentation

- [x] Update react-deep-force-update README
- [x] Update react-hot-loader README
- [x] Update react-stand-in README
- [x] Update main README
- [x] Add changelog for v4
- [x] Update troubleshooting

### Tests

- [x] Add a sanbox environment to test real use-cases
- [x] Add code coverage (codecov)

### Chore

- [x] Add a release scripts to automate release 

### Breaking changes

- [x] Remove webpack plugin

theKashey:
Ok. I’ll take a look. 
I am also trying to solve the last issue about state loss in #703. 
Look like we could repeat and even improve behaviour of v1 (#643)

theKashey:
@neoziro - this is ok, that `modern` tests are failing. The classes in the tests does not `REGENERATE_METHOD`, injected by babel plugin. As result they could not reproduce arrow functions.

neoziro:
@theKashey we have to solve it I think

neoziro:
I think we are now stable on lerna branch. We should add a sandbox projet to be able to test real use-cases and then we could release a beta.

Also would be great to work on a detailed changelog about what's new in v4 and the trade-offs.

theKashey:
So, I could bring back 15/16 boilerplates as sandboxes. And please add one new point - remove webpack plugin.

theKashey:
One more thing to be updated - troubleshooting.

theKashey:
One more thing to be added (and discussed) - `HotExportedComponent`.
```js
 //index.js
 const App = loadable(() => import('./App'))
  
 //App.js
 export default hotExported(module, Component)
```
As long popular loaders, like loadable-components or react-loadable, does not, and from point of view - should not, support ReactHotLoader flow, we might provide a better solution.

V3 allows code change to bubble to the top, and then expects the code to be reinitialized from top to bottom. But loaders will not re-require async components on the fly, and code branches behind them will be kept old. 

V1 had a different behaviour putting module.accept automagically in all the places. That was a quite bad behaviour, and long it has no brains.

Using the helper function user can call HOC in the `right` place with component and module variable. Next RHL can setup module self-acceptance and wrap component with AppContainer to rerender it when the time comes.
The same can be archived for the top level components. Instead of adding
```js
module.hot.accept('./containers/App', () => { render(App) })
```
[...]

Issue #304, opened 2016-05-29
Element type comparing


Mistereo:
It seems like with react-hot-loader@3 it's not possible to check element type for imported components:

``` javascript
const element = <ImportedComponent />
console.log(element.type === ImportedComponent)
```

Logs `false` for me.
Is this a bug or something that should not be supported?

nfcampos:
Hi @Mistereo 

This is intended, react-hot-loader@3 patches `React.createElement`(`<ImportedComponent />` is equivalent to `React.createElement(ImportedComponent)`) so that it returns an element of a proxy wrapper for your components instead of the original component, this is part of allows to replace methods on the component without unmounting.

Do you have any need for `<ImportedComponent />.type === ImportedComponent` to be true?

Mistereo:
@nfcampos I'm trying to implement conditional rendering depending on children types.
I can just set some static prop on these components as a workaround, and check this prop instead.
Just wondering if this can be avoided.

nfcampos:
@Mistereo no, I don't think it can be avoided with the current architecture

JorgenEvens:
I'm running into the same problem, this is how I am working around the issue:

```javascript
const ImportedComponent = (props) => { ... }
const ImportedComponentType = (<ImportedComponentType).type;

const element = <ImportedComponent />
console.log(element.type === ImportedComponentType)
```

singggum3b:
this is definitely a big flaw ! not able to check child type of which class ...

w3apps:
Is it possible to fix this thing or is it too big of a change?

harmony7:
Ran into this today.

One option is to check using `.type.name` like this:

```
import Component from /* ... */;
// Check like this
element.type.name === Component.name
```

However, even this would also break once the resulting code was minified.

So for now I'm using @JorgenEvens 's workaround, thanks for that.

c58:
I used another approach...
```
const ImportedComponent = (props) => { ... }
ImportedComponent.isMyImportedComponent = true;

// Check like this
element.type.isMyImportedComponent === true
```

phyllisstein:
Just ran into this myself and I'm pretty stumped. I'd love to do something a little more elegant than checking `displayName` or cramming my components with static flags, but I can't seem to figure out a solution. Is it really hopeless, @nfcampos ?

Soweyln:
[...]

PR #721, opened 2017-12-18
React-Hot-Loader Tools

theKashey:
Introducing 2 new named exports:
1. `compareComponents`(typeA, typeB) , as was requested in #304. As long getProxy was extracted from patch in reconciler branch - here it actually does nothing.
2. `hotExported`(module, Component), as discussed 2 weeks ago in #707. Also closes #700 

There is no unit tests yet, as I am not ready to introduce karma to test webpack/parsel. But manual testing reports - everything is ok.

Now one can just "mark" export to be _hot exported_, and it will be. As result it will be close to impossible to got a wrong configuration, and it also seamlessly works with any "async-loader"

TODO:
 1. Test with parcel-bundler (it does include [information about RHL](https://parceljs.org/hmr.html), but without example)
 2. Add checks for module exports. As long file is self-accepted, it should not export anything RHL can not handle.
 3. Create `@hotExported` decorator. Not sure hot(in which standard) I can do it.

### DIFF
--- a/packages/react-hot-loader/src/utils.dev.js
+++ b/packages/react-hot-loader/src/utils.dev.js
@@ -1,0 +1,76 @@
+import React, { Component } from 'react'
+import hoistNonReactStatic from 'hoist-non-react-statics'
+import AppContainer from './AppContainer.dev'
+
+const getDisplayName = WrappedComponent =>
+  WrappedComponent.displayName || WrappedComponent.name || 'Component'
+
+const copyReactProp = (source, target) => {
+  hoistNonReactStatic(target, source)
+  target.displayName = `HotExported${getDisplayName(source)}`
+  target.WrappedComponent = source
+
+  target.propTypes = source.contextTypes
+  target.childContextTypes = source.childContextTypes
+  target.contextTypes = source.contextTypes
+}
+
+const makeHotExport = (sourceModule, getInstances) => {
+  const updateInstances = () => {
+    getInstances().forEach(inst => inst.forceUpdate())
+  }
+
+  const thenUpdateInstances = () => Promise.resolve(true).then(updateInstances)
+
+  if (sourceModule.hot) {
+    sourceModule.hot.accept(() => {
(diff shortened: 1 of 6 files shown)


packages/react-hot-loader/src/utils.dev.js
```diff
@@ -0,0 +1,76 @@
...
+        })
+      }
+    }
+  }
+}
+
+export const hotExported = (sourceModule, WrappedComponent, props = {}) => {
+  let instances = []
+
+  class ExportedComponent extends Component {
```
neoziro: We have to [wrap displayName](https://github.com/neoziro/recompact/blob/master/src/wrapDisplayName.js). It is a well known pattern in hoc and it is better for debugging.

packages/react-hot-loader/src/utils.dev.js
```diff
@@ -0,0 +1,76 @@
...
+          if (status === 'apply') {
+            thenUpdateInstances()
+          }
+        })
+      }
+    }
+  }
+}
+
+export const hotExported = (sourceModule, WrappedComponent, props = {}) => {
```
neoziro: The signature is not correct and not composable.

```js
hotExported(module)(WrappedComponent)
```

Also do we really need to specify "module"?

theKashey: The goal is to give RHL control over component's `module`. 
Next it will spinup HRM behavior to enable module self-acceptance, and next wrap Component with AppContainer.
So - this HOCs is doing 2 different things in real, but as long one shall have only one `hot` export per module - it is ok to merge all the stuff inside a single function.

packages/react-hot-loader/src/utils.dev.js
theKashey: It is performed on line 10, but I should use some helper to do it.
PS: And I should not transfer contextTypes. Have no idea why I did it.

PR #722, opened 2017-12-18
Update README for v4

theKashey:
Updating README for RHL and Stand-in.
Text for RHL was made for the `final` version - with tools and reconciler merged.
While updating the code I've spotted what tools or reconciler will solve, and what react-stand-in will break.
Current version could not work without babel plugin. We should make a decision - should RHL operate without babel plugin (TS only), or we can force users to use it in any case.
We need babel only to support arrow functions, but TS could compile into ES2015. Reconciler also relay on the method added by babel plugin to detect "proxyability" of a class, not to wrap anything from node modules.
We still could make it work without babel, but does it worth?

PS: Please keep in mind - I've got a grammar blindness, I cant see mistakes until I'll forget the text.

### DIFF
--- a/packages/react-hot-loader/README.md
+++ b/packages/react-hot-loader/README.md
@@ -49,8 +49,9 @@ module.exports = {
 
 > Note: Make sure to set the `output.publicPath` property to `"/"` as well. Otherwise hot reloading won't work as expected for nested routes.
 
-4. Wrap your application into `<AppContainer>`, all children of `<AppContainer>` will be reloaded when a change occurs:
+4. Setup React-hot-loader to update your application when its code updates.
 
+4.1 Wrap your application into `<AppContainer>`, all children of `<AppContainer>` will be reloaded when a change occurs:
 ```js
 // main.js
 import React from 'react'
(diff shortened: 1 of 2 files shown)


packages/react-hot-loader/README.md
```diff
@@ -49,8 +49,9 @@ module.exports = {
 
 > Note: Make sure to set the `output.publicPath` property to `"/"` as well. Otherwise hot reloading won't work as expected for nested routes.
 
-4. Wrap your application into `<AppContainer>`, all children of `<AppContainer>` will be reloaded when a change occurs:
+4. Setup React-hot-loader to update your application when its code updates.
 
+4.1 Wrap your application into `<AppContainer>`, all children of `<AppContainer>` will be reloaded when a change occurs:
```
neoziro: This part should not appear in this section. It is an advanced case. We have to keep it as simple as possible for the user and the `hotExported` is easier.

packages/react-hot-loader/README.md
```diff
@@ -71,29 +72,45 @@ render(App)
...
+  ReactDOM.render(
+    <Component />,
+    document.getElementById('root'),
+  )
 }
-```
+
+render(App)
+``` 
+This is far more easier way, and it is also __compatible with any source code splitting__ and deferred components loading library,
```
neoziro: To move in advanced section.

PR #723, opened 2017-12-24
docs: rewrite documentation

neoziro:


CHANGELOG.md
```diff
@@ -61,20 +62,21 @@
 * **Breaking change:** instead of
 
   ```js
-<AppContainer component={App} props={{ prop: val }} />
-```
+  <AppContainer component={App} props={{ prop: val }} />
```
theKashey: But it never accepts component. V4 does not change behavior of AppContainer at all.

theKashey: Nevermind.. :0

neoziro: This is old changelog reformated, nothing important.

PR #725, opened 2017-12-24
Trying to fix react 15

theKashey:
Bulk changes to fix all the things.
It still not always working :(

### DIFF
--- a/packages/react-hot-loader/src/AppContainer.dev.js
+++ b/packages/react-hot-loader/src/AppContainer.dev.js
@@ -33,7 +33,7 @@ class AppContainer extends React.Component {
     }
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps() {
     if (this.state.generation !== getGeneration()) {
       // Hot reload is happening.
 
--- a/packages/react-hot-loader/src/AppContainer.dev.js
+++ b/packages/react-hot-loader/src/AppContainer.dev.js
@@ -33,7 +33,7 @@ class AppContainer extends React.Component {
     }
   }
 
-  componentWillReceiveProps() {
+  componentWillReceiveProps(nextProps) {
     if (this.state.generation !== getGeneration()) {
       // Hot reload is happening.
 
@@ -43,7 +43,9 @@ class AppContainer extends React.Component {
       })
 
       // perform sandboxed render to find similarities between new and old code
+      this.renderAnotherChildren = nextProps.children;
       hotReplacementRender(this, hydrate(this))
+      this.renderAnotherChildren = null;
     }
   }
 
@@ -74,7 +76,7 @@ class AppContainer extends React.Component {
       console.error(error)
     }
 
-    return React.Children.only(this.props.children)
+    return React.Children.only(this.renderAnotherChildren || this.props.children)
   }
 }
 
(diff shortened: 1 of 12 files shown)


neoziro:
I think your way to forceUpdate is not correct. We should maybe rely on react-force-update package.

packages/react-hot-loader/src/AppContainer.dev.js
```diff
@@ -43,7 +43,9 @@ class AppContainer extends React.Component {
       })
 
       // perform sandboxed render to find similarities between new and old code
+      this.renderAnotherChildren = nextProps.children;
```
neoziro: What does it fix?

packages/react-hot-loader/src/reconciler/hotReplacementRender.js
```diff
@@ -106,45 +111,49 @@ const mergeInject = (a, b) => {
 }
 
 const hotReplacementRender = (instance, stack) => {
-  const flow = asArray(render(instance))
+  const flow = filterNullArray(asArray(render(instance)))
```
neoziro: This is not perf, we iterate two times. I prefer to check child after.

packages/react-hot-loader/src/reconciler/hotReplacementRender.js
```diff
@@ -106,45 +111,49 @@ const mergeInject = (a, b) => {
...
+    if (typeof child !== 'object' || !child.type || !stackChild.instance) {
       return
     }
+
     if (typeof child.type !== 'function') {
       next(
         // move types from render to the instances of hydrated tree
         mergeInject(
-          child.props ? child.props.children : child.children,
+          filterNullArray(asArray(child.props ? child.props.children : child.children)),
```
neoziro: This is not the fix.

packages/react-hot-loader/src/reconciler/reactUtils.js
```diff
@@ -12,5 +14,7 @@ export const updateInstance = instance => {
     instance.forceUpdate()
   } else if (updater && typeof updater.enqueueForceUpdate === 'function') {
     updater.enqueueForceUpdate(instance)
+  } else {
+    instance._pendingForceUpdate = true
```
neoziro: Useless

packages/react-hot-loader/src/reconciler/reactUtils.js
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable no-underscore-dangle */
 
 export const getReactInstance = instance =>
-  instance._reactInternalFiber || instance._reactInternalInstance
+  instance._reactInternalFiber ||
+  instance._reactInternalInstance ||
+  instance._instance
```
neoziro: OK

neoziro:
Instead of reactHydrate strategy why not walking the React tree, swap what is swappable, mark as pending update. Then walk again and update it?

theKashey:
@neoziro - I've fix close to everything. Tested on pack of local projects - everything is awesome.
Changes:
 1. patch.dev proxies _any_ components. (not only with regenerate method, as before)
 2. reconciler also monkeypatching props during test render. Thus allow to broadcast `not-registered` changes via props (render/children)
 3. Before render - call componentWillUpdate. This is the only way to drop Redux's renderedComponent cache
 4. (Stand-in) After render try to wrap result with proxy. This is the only way to wrap with proxy cached rendered component from (yet again) Redux.

Also - I think few moments could be simplified. Did not test moment about registering the component in `hot` - may be it is not necessary (but it could).

PR #727, opened 2017-12-25
Fix decorators

neoziro:
Running example on 4.0.0-beta.3, I got this:

screen shot 2017-12-25 at 11 52 22


theKashey:
No. __Exactly this one__ will not work.

autobind is a bit straightforward thing - when you access class method it binds it to current `this` and sets as a member of `this`.
This may works... unless stand-it will access prototype. autobind is so straightforward, that on property read it will bound all methods to the current `this`, and `this` will a prototype object, I am ready property from.

As result - `this` inside increment is a class's prototype itself :)

First part is solvable. I could use another, more safer way to iterate thought prototype (same as autobind uses), but in any case - 

`autobind will seal a whole class. There is no way to hot-reload it.`
It creates descriptors in the current context, they are not swappable by prototype change. And nobody could programmatically remove them.

To solve this case we need react-proxy. But it cant stand arrow functions....

PS: One should check getters/setters.

theKashey:
So, I'll take this task, and will add some fixes to make application at least just work, even without reloading.
RHL might not work, but shall not blow the application.

packages/react-stand-in/src/inject.js
```diff
@@ -61,6 +64,9 @@ function mergeComponents(
               key,
               nextAttr,
               '. Unable to reproduce, use arrow functions instead.',
+              `(arity: ${nextAttr.length}/${prevAttr.length}, proto: ${
```
theKashey: Temporary for debug.

packages/react-stand-in/src/inject.js
```diff
@@ -98,18 +104,29 @@ function mergeComponents(
...
+}
+
 function checkLifeCycleMethods(ProxyComponent, NextComponent) {
   try {
-    const p1 = ProxyComponent.prototype
+    const p1 = Object.getPrototypeOf(ProxyComponent.prototype)
     const p2 = NextComponent.prototype
     reactLifeCycleMethods.forEach(key => {
-      if (String(p1[key]) !== String(p2[key])) {
+      const d1 = Object.getOwnPropertyDescriptor(p1, key) || { value: p1[key] }
```
theKashey: Do not access prototype directly, but try to do it via descriptors

theKashey:
Ok. This changes will fix issue with `autobind`. The solution is not to access prototype directly, but use property descriptors for it.
At the end - hot reloading actually does work, as long we also copy over descriptors on proxy update.

Issue #738, opened 2017-12-27
Handle async loading of patch


neoziro:
As of now, if a module is registered before `react-hot-loader/patch` is loaded, it is ignored. The new setup of React Hot Loader v4 does not require to load `react-hot-loader/patch` before everything in Webpack.

https://github.com/gaearon/react-hot-loader/blob/41520935adcf655aab033b28e1c660f62be494ae/packages/react-hot-loader/src/babel.js#L30-L40

We need a strategy to being able to call `__REACT_HOT_LOADER__.register` without having `__REACT_HOT_LOADER__`.

The strategy consists by creating a temporary `__REACT_HOT_LOADER__` directly in Babel if not available and to catch up when `patch.js` will be loaded. This is exactly the same [strategy applied by Google Analytics with global `ga`](https://developers.google.com/analytics/devguides/collection/analyticsjs/).

PR #739, opened 2017-12-27
fix: handle async loading of patch

neoziro:
Implement #738

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,13 @@ module.exports = {
   rules: {
     'no-underscore-dangle': [
       'error',
-      { allow: ['__REACT_HOT_LOADER__', '__standin_getCurrent'] },
+      {
+        allow: [
+          '__REACT_HOT_LOADER__',
+          '__standin_getCurrent',
+          '__registeredComponents',
+        ],
+      },
     ],
     'no-console': ['error', { allow: ['warn', 'error'] }],
     'no-plusplus': 'off',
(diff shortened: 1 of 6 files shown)


theKashey:
Hey! I've simplified babel. Look like it works well.

packages/react-hot-loader/src/patch.prod.js
```diff
@@ -1 +1,4 @@
 /* noop */
+module.exports = {
+  default: false,
```
neoziro: `null`

.eslintrc.js
```diff
@@ -4,7 +4,13 @@ module.exports = {
   rules: {
     'no-underscore-dangle': [
       'error',
-      { allow: ['__REACT_HOT_LOADER__', '__standin_getCurrent'] },
+      {
+        allow: [
+          '__REACT_HOT_LOADER__',
+          '__standin_getCurrent',
+          '__registeredComponents',
```
neoziro: Not needed.

packages/react-hot-loader/src/babel.js
```diff
@@ -31,10 +31,11 @@ module.exports = function plugin(args) {
   const buildTagger = template(
     `
   var UNUSED = (function () {
-    if (typeof __REACT_HOT_LOADER__ === 'undefined') {
-      return;
+    var __REACT_HOT_LOADER__ = global.__REACT_HOT_LOADER__ || require('react-hot-loader/patch').default;
```
neoziro: Global does not work in browser.

Issue #732, opened 2017-12-27
Replace warnings by debug and defaults it to false


neoziro:
We have a [prop `warnings`](https://github.com/gaearon/react-hot-loader/blob/next/packages/react-hot-loader/src/AppContainer.dev.js#L97) that activate some warnings. It is only relevant for advanced users and for debug purpose. The goal is to rename it everywhere with `debug`, defaults it to `false` and document it in README.

**Be careful by replacing every `warnings` present in code.**

theKashey:
`debug` or `dev`?
How to pass options without AppContainer?
```js
import {enableDebug} from 'react-hot-loader';
enableDebug();

//or
import 'react-hot-loader/debug'
```

neoziro:
Actually we can do it using `hot(module, { debug: true })`. What do you think?

neoziro:
`dev` is confusing because we are in "development" environment when we use RHL.

theKashey:
How many `hot` exports you may have in your project?
As I expected - at least one for every for every entry point (including chunks).
As I understand - you may have only one, at _loadables_ will still works
(PS: this mean than we could monkey patch `react-dom`)

But anyway - as long you may have more that one `hot` - this global setting should not be a part of it.

neoziro:
@theKashey I agree, do you think we will need other options? I think about:

```js
import { configure } from 'react-hot-loader'
configure({ debug: true })
```

theKashey:
To say the truth - it should not have options. But `configure` might be the simplest way.

theKashey:
Not finished yet.

neoziro:
@theKashey OK, waiting for PR.

neoziro:
#743 will definitely fix this one

PR #740, opened 2017-12-27
feat: replace warnings by `configure({ debug: true })`

neoziro:
Implements #732


neoziro:
@theKashey OK, I merge it, I let you continue

PR #744, opened 2017-12-28
Fix componentWillUpdate error

katmai7:
Hi, 
I have tried to move to v4 version and faced with an error(`nextProps/nextState is undefined`), when componentWillUpdate is [called](https://github.com/gaearon/react-hot-loader/blob/next/packages/react-hot-loader/src/reconciler/hotReplacementRender.js#L98) on the non react-redux connected component.

As a solution: I suggest to ensure that the component is a connected component and after that call the componentWillUpdate.

As a second solution: pass nextProps and nextState to the componentWillUpdate, but I'm not sure that it is possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 lib
 coverage
 .DS_Store
+.idea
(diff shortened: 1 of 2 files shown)


neoziro:
Thank you for reporting and trying to fix it! The second solution is better, I did it in #747.

neoziro:
@katmai7 I didn't ask you to do it because I was not sure how to do it exactly (especially for test).

PR #746, opened 2017-12-28
docs(example): add styled-components / emotion example

neoziro:
@theKashey there is a problem:

When we edit `<SmallText>Hello world!</SmallText>` we got a warning then component is fully replaced (we only have counter displayed).


theKashey:
This is release blocker.

Issues:
1. createElement got not replaces in App.js. It is actually createElementWithValidation. Will work only if {hot} will be required FIRST.
The root issue is quite-quite funny:
```js
import * as React from 'react'
// produces React.createElement after babel <-- got not replaced.
import React from 'react'
// produces React.default.createElement after babel
```
Solution - import in this manner in patch.dev, and patch all the ways...

2. Styled-components seal props. Everything like `ProxyComponent.RHL_PROXY_NAME` will have value "undefined"

Solution - use defineProps for all our props. Make #748 obsolete

3. Styled-components do have more that one parent.
Standin does replace only the first one. And the second will throw error on construction.

Solution - update all parents in prototype stack...

I've spiked all the things, and do have a working solution. Will convert to a normal PR asap.

theKashey:
Last change makes #748 obsolete and fixes any issue with own props. 
Including the bug we had in `transferStaticProps` - it actually removes all props added by standin.

Styled component case also works well, with proper config of course. Without proper config it will throw an error.

packages/react-hot-loader/src/reconciler/hotReplacementRender.js
```diff
@@ -199,14 +199,24 @@ const hotReplacementRender = (instance, stack) => {
     } else {
       // unwrap proxy
       const childType = getElementType(child)
+      if (!stackChild.type[PROXY_KEY]) {
```
theKashey: Trespasser!!

packages/react-hot-loader/src/reconciler/hotReplacementRender.js
```diff
@@ -199,14 +199,24 @@ const hotReplacementRender = (instance, stack) => {
...
+        /* eslint-disable no-console */
+        console.log(
```
neoziro: We should use logger.

packages/react-stand-in/test/consistency.test.js
```diff
@@ -173,6 +182,20 @@ describe('consistency', () => {
...
+        const BarProxy = barProxy.get()
+        expect(BarProxy.toString()).toMatch(/^(class|function) Bar/)
+
+        Object.defineProperty(Foo, 'toString', {
+          configurable: false,
+          value: Foo.toString,
+          writable: false,
+        })
+        createProxy(Foo, 'foo')
+        expect(logger.warn).not.toHaveBeenCalledWith()
```
neoziro: We should remove it.

Issue #758, opened 2017-12-30
[4.0.0-beta.11] Untranspiled RHL code in production bundle


apostolos:
### Description

Production bundle includes untraspiled *.prod.js code. Among other things, this breaks Uglify minifaction step.

### Expected behavior

Stub code should be transpiled.

### Actual behavior

The content of the following files is included "as is" on the **main** webpack chunk:

- https://github.com/gaearon/react-hot-loader/blob/next/packages/react-hot-loader/src/hot.prod.js
- https://github.com/gaearon/react-hot-loader/blob/next/packages/react-hot-loader/src/AppContainer.prod.js
- https://github.com/gaearon/react-hot-loader/blob/next/packages/react-hot-loader/src/utils.prod.js

### Environment

React Hot Loader version: 4.0.0-beta.11
Babel: 7.0.0-beta.36
Webpack: 4.0.0-alpha.2

### Reproducible Demo

https://github.com/LWJGL/lwjgl3-www

theKashey:
Thank you. Will fix asap.

SuhairZain:
@theKashey Hey,
Is there an update on this? We were looking to migrate from v3 to v4 because of an issue with setState. But doing so would break our production build with uglify.

theKashey:
PR already opened. Waiting for completion.

neoziro:
Fixed in [v4.0.0-beta.12](https://github.com/gaearon/react-hot-loader/releases/tag/v4.0.0-beta.12)

apostolos:
Many thanks!

RHL v4 now in production here: https://www.lwjgl.org/

PR #759, opened 2017-12-30
Use es5 code for prod. #758

theKashey:


neoziro:
Why?

theKashey:
We do not transpile code to es5 cos we have to use es6 classes to work with es6 classes. 
Not sure it will be easy to use two separate Babel configs for this task.

neoziro:
@theKashey please let me do it, build is my kind of stuff

neoziro:
Continue in #761.

PR #761, opened 2018-01-01
 fix: target ES5 for production code

neoziro:
Following #759.


theKashey:
CI stuck, and I even cant stop it. But it seems to be ok. Merge?

PR #769, opened 2018-01-06
`module.hot` callback should indicate to use the latest `component`

ryuever:


neoziro:
@ryuever Thanks but we are about to release [v4](https://github.com/gaearon/react-hot-loader/tree/next). We do not accept any changes for current version.

Issue #775, opened 2018-01-10
react-stand-in incorrectly picks up Relay containers as functional components


taion:
This is against v4 beta 13.

Relay Classic uses an odd pattern to lazily build containers: https://github.com/facebook/relay/blob/d77a34c6b5abddfe078b0d50c2d43181fe35d574/packages/react-relay/classic/container/RelayContainer.js#L1050-L1055

This behaves as a normal class, but does not get picked up as a React class here.

theKashey:
Ok, so it uses a bit uncommon feature, about returning another class instance from a constructor.

taion:
Yeah – let me see if I can monkeypatch my way out of this. This is sort of a stupid edge case anyway.

theKashey:
It is not a simple to fix this stuff, as long you will have to `hot-patch` a Component returned from RelayContainer, which may be tricky with ES6 classes.
But we can not wrap such constructors at all, just to not break anything.

taion:
This appears to work:

```js
import Relay from 'react-relay';

if (__DEV__) {
  const createRelayContainer = Relay.createContainer;

  Relay.createContainer = (...args) => {
    const Container = createRelayContainer(...args);
    Container.prototype.isReactComponent = {};

    return Container;
  };
}
```

Can you reliably distinguish such constructors from normal functional components though? They seem to look much the same from the outside.

aminland:
@taion can you try this: #776 

I spent the better part of the last 6 hours trying to figure out wtf was going on lol... It seems to fix it for me...

theKashey:
@aminland - no, that's not a solution.
I am looking forward to create `isComponentWrappable` function, if not - just dont proxy.
We actually had that logic before, and dont wrap anything inside node_modules, but next decide that it is doable.
It is doable expect returning another instance from a constructor.
So
```js
const isComponentWrappable (BaseClass) => {
  return testES5(BaseClass) || testES6(BaseClass);
}
const testES5 = (baseClass) => {
  function Tester (){
     const that = baseClass.call(this);
     return that==this;
  }  
  try{
    return !!(new Tester());
  } catch(e) {
    return false; // it was ES6 class
  }
}

const testES6 = (baseClass) => {
   class Tester extends baseClass {
      constructor(props) {
         return super(props);
      }
      MAGIC_FUNCTION(){}
   }
  try{
    return !!(new Tester()). MAGIC_FUNCTION
  } catch(e){
    return false;
  }
} 
```
[...]

PR #776, opened 2018-01-10
Fix for proxying issue with react-relay

aminland:
Using react-relay, I was getting an error that occurred because a relay component was missing its $$typeof property. 

Relay's createFragmentContainer and createRefetchContainer functions don't actually wrap the original item at all, but rather return a new object which has it's own render method (but is not a valid element for mounting). 

I'm sure there's a cleaner fix, but couldn't think of a good one just yet...

### DIFF
--- a/packages/react-stand-in/src/createClassProxy.js
+++ b/packages/react-stand-in/src/createClassProxy.js
@@ -63,10 +63,13 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
     render() {
       inject(this, proxyGeneration, injectedMembers)
 
-      const result = isFunctionalComponent
+      let result = isFunctionalComponent
         ? CurrentComponent(this.props, this.context)
         : CurrentComponent.prototype.render.call(this)
-
+      
+      if (!!result && !result.$$typeof && result.render){
+        result = result.render()
+      }
       return wrapResult(result)
     }
   }


aminland:
Didn't notice someone made an issue for this already (#775 )

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -63,10 +63,13 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
...
       inject(this, proxyGeneration, injectedMembers)
 
-      const result = isFunctionalComponent
+      let result = isFunctionalComponent
         ? CurrentComponent(this.props, this.context)
         : CurrentComponent.prototype.render.call(this)
-
+      
+      if (!!result && !result.$$typeof && result.render){
+        result = result.render()
```
taion: I'm not sure this is right. I don't think we should e.g. be creating a new Relay component instance every time in `render`. And anyway this should use `this` as context as above.

taion:
This might well work, but I'm not totally sure it's correct. This is essentially using a new component instance for every render. It seems inefficient and the semantics aren't quite correct.

Is there any way to figure out that these things aren't really functional components, and are instead just weird class components?

taion:
Alternatively, I'm not sure anybody other than Relay Classic uses this pattern, so you could just sniff for `getFragment` in `isReactClass`.

packages/react-stand-in/src/createClassProxy.js
theKashey: We cant place such `secret-knowledge` based logic here.
As long it may fix Relay - it can break something else. Or not fix something else.

taion: This isn't right though. \:

Constructing a new instance of the component is unlikely to work in the general case.

Maybe just throw an error here and tell the user to install their own monkeypatch per https://github.com/gaearon/react-hot-loader/issues/775#issuecomment-356508850?

theKashey: We dont have any reasons to throw an error. I'll expect a lot of false positives here.
Just give us a bit of time to find a proper way to solve this case as a broad problem, not bound to Relay case.

theKashey:
Obsolete due to #778, big kudos for effort.

PR #778, opened 2018-01-10
Support IndeterminateComponents

theKashey:
This will solve #775.
The logic is simple - use the same detection mechanism as react's `mountIndeterminateComponent` has to detect `class` returned from a render function.
In case of such object - wrapp result with StatelessFunctional component and __preventing__ RHL to proxy it.
As result - this will re-enable native React logic to handle this case.

Drawback: you could return such thing from any component, including Stateful Classes, as long you could not do it in unpatched React. Maybe we should use this trick only for functional components.


packages/react-hot-loader/src/reactHotLoader.js
```diff
@@ -9,7 +10,7 @@ import {
 } from './reconciler/proxies'
 
 function resolveType(type) {
-  if (!isCompositeComponent(type)) return type
+  if (!isCompositeComponent(type) || type[DO_NOT_PROXY_KEY]) return type
```
theKashey: To let StatlessFunctionComponent exists.

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -1,16 +1,32 @@
...
   identity,
   safeDefineProperty,
 } from './utils'
 import { inject, checkLifeCycleMethods, mergeComponents } from './inject'
 
 const proxies = new WeakMap()
 
+function Dereference(props) {
+  // copy over pros
+  const classInstance = props.standin_classInstance
```
theKashey: returned instance must contain the same props, or will trigger a warning.

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -67,6 +83,10 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
         ? CurrentComponent(this.props, this.context)
         : CurrentComponent.prototype.render.call(this)
 
+      if (!isReactElement(result)) {
```
theKashey: The actual fix.

theKashey:
Tests to be added a bit later.

packages/react-stand-in/src/createClassProxy.js
taion: I'm not sure this is right. Per https://github.com/facebook/react/blob/e6e393b9c5221bfb1a5ddcc7221c42e96ab3baca/packages/react-reconciler/src/ReactFiberBeginWork.js#L446-L450, the "indeterminate component" logic seems to only get run for the initial mount.

This logic will actually construct a new component instance on every render call, which doesn't seem like what we want.

We shouldn't be creating new class instances if we can help it at all.

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -1,16 +1,32 @@
...
 
+function Dereference(props) {
+  // copy over pros
+  const classInstance = props.standin_classInstance
+  classInstance.props = props
+  return classInstance
+}
+Dereference.DO_NOT_HOT_RELOAD = true
+
+function wrapWithStateless(classInstance, props) {
```
taion: I had to read this a few times to figure out what this was doing. It's not exactly wrapping this with a stateless functional component – it's more like `rewrapAsIndeterminate`.

packages/react-stand-in/src/createClassProxy.js
taion: Maybe the answer is to cache `result` in this case? It shouldn't invalidate, I don't think.

packages/react-stand-in/src/createClassProxy.js
theKashey: It just creates `a real` functional component, which will be Interminate until the execution, and can carry desired payload.

packages/react-stand-in/src/createClassProxy.js
theKashey: The logic is the simple - `how it will work __without__ RHL?`
Without RHL it will construct and return a new Component on each render call, and will not _cache_ anything.
That's the answer.

We can do a lot, but we shouldn't.

taion: Are you sure about that? The snippet above has this check:

```js
    invariant(
      current === null,
      'An indeterminate component should never have mounted. This error is ' +
        'likely caused by a bug in React. Please file an issue.',
    );
```

So I believe this logic only gets applied at initial mount?

theKashey: There is no such thing as `initial` mount for a function stateless component. As I understand - it just can't `change its mind`.
In real - this code works without any exceptions or warnings.

packages/react-stand-in/src/createClassProxy.js
taion: This isn't a standard functional component, though. In this case, the return value is not a React element – it's a React component instance.

As such, it goes through that "indeterminate" code path, and React realizes that this is actually a class component.

packages/react-stand-in/src/createClassProxy.js
aminland: Random question: why can't we just wrap functional components separately (not as a react component), and then _only_ create a proxy for it's results if it generates a valid react component? 
If it's returning an element or anything else we don't really need to care. This would also be much simpler, and probably add less overhead...

packages/react-stand-in/src/createClassProxy.js
theKashey: Yet again - just remove RHL from you mind.
1. Reproduce the problem.
 You will found that you can return renderable-component only from Stateless functional, not from class-based. (and there is no third way).
2. Let RHL to mimic same behaviour.
 RHL removes functional components, making all class-based - thats is the problem. We have to do it, cos this is the only(*) way to be able hot-swap function implementation.

This means - if you found some odd behavior, you could reproduce only using functional component - use functional component to reproduce the original behavior. 
This also means - we should wrap functional components with functions. It will be the best way to solve any further issues.

theKashey:
! Not to be merged yet.

packages/react-stand-in/src/createClassProxy.js
aminland: +1 on that last point, probably the only way that will work in most cases (save for a function storing state in its scope or at module level)

packages/react-stand-in/src/createClassProxy.js
theKashey: Cos first you have to wrap them, next you can test something.

packages/react-stand-in/src/createClassProxy.js
theKashey: Just spiked possible solution with keeping functions as functions - it is much more clear, but results 5% tests to fail.
I'll hope to work it out till the end of the day.

taion: @theKashey 

Take a look at this CodeSandbox: https://codesandbox.io/s/6wz0or0rqk

This is not exactly a stateless functional component. It gets resolved as a normal class-based component.

In the first set of commands, if you look at the console output, you'll see `FooConstructor` is only called once, while `Foo.render` is called 3 times. If you look at https://6wz0or0rqk.codesandbox.io/ with React Dev Tools, you'll see that a single `FooConstructor` component gets mounted, and it is a full-fledged class component with state and everything.

By contrast, if you uncomment out the render calls with the proxy, you'll see that:
- The component tree is different
- `FooConstructor` gets called 3 times instead of just 1 time

So it's not the same. But in both these cases, no stateless functional components are getting rendered.

taion: i.e.

image

For

```js
import React from 'react';
import ReactDOM from 'react-dom';

class Foo extends React.Component {
  constructor(props, context) {
    super(props, context);

    this.state = {
      foo: 'bar',
    };
  }

  render() {
    console.log('foo');
    return <div>foo</div>;
  }
}

function FooConstructor(props, context) {
  console.log('foo constructor');
  return new Foo(props, context);
}

ReactDOM.render(<FooConstructor />, document.getElementById('root'));
ReactDOM.render(<FooConstructor />, document.getElementById('root'));
ReactDOM.render(<FooConstructor />, document.getElementById('root'));
```

There are no stateless functional components involved here.

theKashey: Unfortunately, we have to proceed with this approach.
I've spiked the other way to wrapping the stuff, keeping functional components functional, and works with this case without any magic, but....
But stateless component does not leave `instance` behind, preventing _hot-reconciler_ to properly work. 

To fix this one have to rework `getReactStack` to properly handle any possible React.Element type.

theKashey:
PR is feature complete. 
 - Fixes an issue with IE11 (RHL still may not work due network/HMR)
 - Mitigates an issue with Indeterminate Component

PR to follow - reactStack should support native functional component. Currently resulting empty instance, then stopping reconciler to dive into.

theKashey:
There is one change for @neoziro - we dont need ES6/2015 environments any more.
To handle IE11 case we have to compile __all__ the code to es5, and construct `class` for standin using the `eval`.

taion:
@theKashey 

Do you have code up for https://github.com/gaearon/react-hot-loader/pull/778#discussion_r160875826?

I wonder if it'd be possible to use the same indeterminate trick here. In the functional case, return a function that returns the class if `InitialComponent` is functional.

Easier to show with code than words.

taion:
Something like:

```js
let ProxyComponent;

if (isFunctionalComponent) {
  ProxyComponent = (props, context) => {
    // This will only get called for the initial mount.
    const result = InitialComponent(props, context);
    if (!React.isValidElement(result)) {
      // This is something weird like a Relay container, where a non-class
      // constructor returns a component instance. We can't proxy this, so just
      // pass it through as-is.
      return result;
    }
    
    return proxyFunctionalComponent(InitialComponent);
  };
}
```

theKashey:
@taion 
```js
if (isFunctionalComponent) {
    ProxyComponent = function(props, context) {
      return wrapResult(CurrentComponent(props, context))
    }
  }
.....
```
Thats all. But we have to stick to Classes, as long we need `instances` to make reconciler work.

taion:
@theKashey 

Due to time zone difference I can't code this up right now, but I'll try something out tomorrow. I think `ProxyComponent` can use the same trick as in `ContainerConstructor` where it can be a function returns a component instance.

As in the sandbox I linked, for all intents and purposes, this behaves the same as if you used a class component in the first place. It just provides an extra hook to e.g. return a component instance in something like the Relay case.

theKashey:
@taion - the correct way to solve this - find another way to hydrate React tree, supporting functional components. But I am not sure about this right now.

"IndeterminateComponent" is a quite strange thing - it executes only one, dissolves in ReactTree, and next you __can't__ change its result.
It memoizes the component it returned, and freezes result till unmount. There is no way to `hot-reload` returned instance.
RHL is not working for this component class, due to React native behavior. And we should not fight with React.

packages/react-stand-in/src/createClassProxy.js
aminland: See this sandbox: https://codesandbox.io/s/jl3267871y

theKashey: During the hot replace we start process to merge the old and the new react tree, to update components which are not exposed as variables or just hidden behind decorators. 
During this process we can’t handle function components, except this case, as long React stores only return variable in the tree. 
Your code example is eligible, but hotReplacement can’t pass though the real functional component. 

For now - everything except this case have to be wrapped with class.

theKashey:
Anyway - lets step back, and try to improve `getReactStack` to properly handle functional components.
We got too many new magic here.

taion:
@theKashey Can you clarify what you mean when you say "functional components" here? In the case here, there are no functional components involved. There are only stateful class components.

theKashey:
The problem is - we are converting `function(a) { return b;}` to `class A{ render() {return b;}}`. But there is some React specific, so classes and function are not equal.
As result - _There are only stateful class components._, as you said :)
If we will stop doing it - it will also solve the issue, and make the code a bit more clear.

taion:
@theKashey I opened https://github.com/gaearon/react-hot-loader/pull/785 with an implementation of https://github.com/gaearon/react-hot-loader/pull/778#issuecomment-356845557. This is similar to the current implementation, but avoids the extra wrapper component instance in the indeterminate case. It comes at the cost of calling stateless functional components twice on the initial render, but the code is a lot simpler, and stateless functional components ought to be pure anyway.

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -40,18 +40,14 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
...
   function proxiedRender() {
     inject(this, proxyGeneration, injectedMembers)
 
     let result
 
-    if (hasInitialResult) {
-      result = initialResult
-      hasInitialResult = false
-      initialResult = null
+    if (this[CACHED_RESULT]) {
```
taion: You may need to do an explicit `hasOwnProperty` check here. Imagine if the functional component returns `null` or something – then this check will fail. That's why I had the explicit `hasInitialResult` earlier.

theKashey: Nope. Result have to be an instance with a render function to trigger this code branch.

taion: Sorry, I may not have explained correctly in the comments. This code path only gets hit for normal functional components. When the proxies component actually returns a React instance, we don’t hit this code path. The initial result is instead the initial return value from a normal functional component.

taion: See https://github.com/gaearon/react-hot-loader/commit/f27db13ffbfd8f2ab753b94a6384a2b21ca55146. This affected the test results for SFCs.

theKashey: You are right. I've `inverted` the expectations.

taion: I can PR a fix and add a regression test.

theKashey: It will be awesome. I’ve stuck with home duties.

PR #783, opened 2018-01-11
Add a PWA boilerplate to the list

alexnoz:


theKashey:
I've changed the target or this PR to the proper branch - "next", but it will not gonna work :(
We are not accepting changes for version 3, could you please raise another PR for the version 4?

PR #785, opened 2018-01-12
Use indeterminate path for functional components

taion:


packages/react-hot-loader/src/reconciler/hotReplacementRender.js
```diff
@@ -168,8 +167,7 @@ const hotReplacementRender = (instance, stack) => {
       // copy over props as long new component may be hidden inside them
       // child does not have all props, as long some of them can be calculated on componentMount.
       const nextProps = { ...instance.props }
-      const childProps = child.props || {}
```
taion: I'm not sure if this check was related to the indeterminate component handling, or if it was there for some other reason. Tests seem to pass without it.

packages/react-hot-loader/test/AppContainer.dev.test.js
```diff
@@ -968,7 +968,7 @@ describe(`AppContainer (dev)`, () => {
       )
       expect(wrapper.find('App').length).toBe(1)
       expect(wrapper.contains(<div>hey</div>)).toBe(true)
-      expect(spy).toHaveBeenCalledTimes(1)
+      expect(spy).toHaveBeenCalledTimes(2)
```
taion: As noted, we end up calling functional components twice now. We could cache the initial result if you want, though.

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -48,12 +47,42 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
...
+    // rendered component instance will be the return value.
+    ProxyComponent = function(props, context) {
+      const result = CurrentComponent(props, context)
+
+      // This is a Relay-style container constructor. We can't do the prototype-
+      // style wrapping for this as we do elsewhere, so just we just pass it
+      // through as-is.
+      if (isReactIndeterminateResult(result)) {
+        ActualProxyComponent = null
+        return result
```
taion: Technically we could use the older ways of proxying this here. Maybe this should be a `FIXME` or a `TODO`. I've never seen this pattern get used outside of Relay, and you can't really hot-reload Relay containers in a useful way anyway, because the queries are static.

taion:
I added the caching as described to avoid calling a functional component twice on the initial render. The tests now pass as-is, except the no-op call in the indeterminate component case is gone.

theKashey:
Much less magic! LGFM.
Anyway - let me test some edge cases

packages/react-hot-loader/test/AppContainer.dev.test.js
```diff
@@ -1340,7 +1340,7 @@ describe(`AppContainer (dev)`, () => {
 
         // How it works. IndeterminateComponent calculates return value once.
         expect(wrapper.text()).toBe('hey 44 new')
-        expect(spy).toHaveBeenCalledTimes(1) // called during reconciler, has no effect on result
+        expect(spy).toHaveBeenCalledTimes(0) // never gets called
```
theKashey: Not it mimics React behaviour better.

taion: yeah... it's just a laziness thing. i think one of the old versions of RHL had a proxy strategy that worked on instances (maybe?), but it's not really worth it since relay queries are static anyway, and nobody else uses this pattern

theKashey: Why somebody even invents it!? 🔮

taion: I believe it’s because Relay containers are relatively expensive to create, and idiomatic Relay code puts a container around nearly every component. Given that, my guess is that it’s an optimization to improve startup time. 

I think that’s also why nobody else uses this pattern, and why we don’t really have to worry about it. Unless you’re building static container classes everywhere, there’s no point to using it.

taion: A container class around every component class, I mean. In Relay, nearly every component is a smart component, sort of.

aminland: Can this be merged and released then?

taion: Note that this is against the `indeterminateComponent` feature branch, not `next`.

Until this gets released, you can still use this monkeypatch, though you'd need to do a bit more complicated stuff with modules for modern/compat than for classic/legacy: https://github.com/gaearon/react-hot-loader/issues/775#issuecomment-356508850.

theKashey:
I’ll handle this soon(sorry, time difference).

PR #788, opened 2018-01-14
Handle falsy initial SFC render result

taion:


packages/react-stand-in/src/utils.js
```diff
@@ -78,7 +78,7 @@ const ES5ProxyComponentFactory = function(
   return ProxyComponent
 }
 
-export const isReactIndeterminateResult = el =>
+export const isReactComponentInstance = el =>
```
taion: This is a slightly clearer name, I think, since it's what we're actually checking for here.

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -82,20 +85,23 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
...
+      if (isReactComponentInstance(result)) {
         ProxyComponent = null
         return result
       }
 
       // Otherwise, it's a normal functional component. Build the real proxy
       // and use it going forward.
       ProxyComponent = proxyClassCreator(Component, postConstructionAction)
+
+      // We don't need to use safeDefineProperty here, as we control the entire
```
taion: I hope this is correct. Otherwise we should change this to use `safeDefineProperty` as well.

packages/react-hot-loader/test/AppContainer.dev.test.js
```diff
@@ -971,6 +971,23 @@ describe(`AppContainer (dev)`, () => {
...
+      }
+      RHL.register(App, 'App', 'test.js')
+
+      const wrapper = mount(
+        <AppContainer>
+          <App />
+        </AppContainer>,
+      )
+      expect(wrapper.find('App').length).toBe(1)
+      expect(spy).toHaveBeenCalledTimes(1)
```
taion: This test looks silly, but it does in fact catch the regression.

PR #793, opened 2018-01-15
Update troubleshooting document with clientLogLevel information

donaldtf:
@theKashey Here is the new pull request for the 'next' branch. And the link to my outdated one for posterity. https://github.com/gaearon/react-hot-loader/pull/791#issuecomment-357569660

Thanks!


theKashey:
The only problem now - "React Hot Boilerplate" is uber outdated.

PR #802, opened 2018-01-18
A better HRM 

theKashey:
1. Auto require updated module in `hot`, as long `Electron` does not do it by itself.
- has no effect for webpack (already required)
- does not work for parcel (no module.id)

2. Add notes how to setup HRM.
 - do not use `hot` for parcel bundler. (it does not work, and our example also does not work)
 - how to setup electron
 - be just accurate :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ npm install react-hot-loader@next
 }
 ```
 
+> Note: use `.compilerc` in case of Electron
+
 2. Mark your root component as _hot-exported_:
 
 ```js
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,6 +35,8 @@ npm install react-hot-loader@next
 }
 ```
 
+> Note: use `.compilerc` in case of Electron
```
neoziro: I think we should not add note in this section.

README.md
```diff
@@ -47,6 +49,8 @@ const App = () => <div>Hello World!</div>
 export default hot(module)(App)
 ```
 
+Do not use `hot` if you are using **parcel** bundler. It was designed for webpack.
```
neoziro: We should make it work on Parcel too.

neoziro: Can you explain what is the problem? I think we should open a new issue on Parcel, if it does not work with this project, it is useless.

theKashey: https://github.com/parcel-bundler/parcel/issues/619
Meanwhile, I've found a bug in `hot`. From now `hot` will throw an error is `.id` is not defined along the module.

theKashey: Follow up - https://github.com/gaearon/react-hot-loader/pull/818

PR #810, opened 2018-01-21
Fix running tests on Windows

insin:
- Use `cross-env` to set environment variables
- Replace backslashes in the `__FILENAME__` replacement `RegExp`


neoziro:
Thanks!

PR #814, opened 2018-01-22
Swallow the error during the hot-render phase.

theKashey:
try...finally does not catch the error, one should `catch` it in real.

As result "the error" before and after HRM has the same behavior. We do not changing it now.


neoziro:
Coverage is decreasing, can you add a little test for this?

neoziro:
Hmm, and why not throw it instead of logging?

theKashey:
Cos error should be thrown by real render. And it should be thrown in the real render.

Issue #806, opened 2018-01-19
Get rid of AppContainer


theKashey:
Currently, we are asking to wrap some part of an Application with <AppContainer />.
Prior v4 it was used to "deep-force-update" the rendering tree. In v4 the goal of AppContainer is to perform `hot-render` to understand how application updated, and thus update some hidden behind HOC and decorators stuff.

Hot loader was a quite simple thing in the beginning, but day by day, case by case, it becomes more complex and implements more and more React details, it should not know about.

The problem is simply - hot-loader has to be more simply. And if it will be simply - it will fail to reconcile some Applications.

One could wrap any component at any level with `hot` to "scope" an update to a small branch, and help hot-render not-to-face complex cases.

However, nothing stops up from integration AppContainer into __all__ components. Just before the render (as long we control this method) one could double check that render was not called from existing `hot-render`, and then could run it's own hot-render.

As result - if something prevents AppContainer's hot-render to reconcile deeper nodes, they will run reconciliation by themselves.

Outcome - we could remove 50% code from hot-renderer and still be able to handle any case. Especially crazy stuff about children and fragments.
More of it - this should better handle portals (currently we don't handle them), and maybe future async render.

Performance hint: a single variable comparison on render, a single variable to be stored among instance.

neoziro:
OK so you mean transforming `const Hello => 'hello'` into `const Hello => <Hot>hello</Hot>`, is it right?

I think it is not possible because we have no idea if it is a simple function or a component.

theKashey:
No, I mean add one line to a ProxyComponent's `render` method.

Currently, it checks that elements "version" is the same as the last one. If not - injects the last version. Ie variables and bound functions. Prototype is already overriden.

Afterwards, will also check that elements was _rendered_ as a new version (or not rendered before). If not - call hotReplacementRender with itself as root.

Make any component behave as AppContainer.

neoziro:
@theKashey OK I understand.

theKashey:
The only feature we can achieve here - handle portals.
The only improvement we can get from the change - simplify the code.
The only concern - @gaearon said that he is going to work on hot-reloading in this year. Should we introduce a new level of complexity(even if it is simpler) hacking deeper into React, but just wait for a bright future, when all these features will be a part of react-dom reconciler.

theKashey:
So - I'll try to handle this idea on weekends. And look like it's a "nothing more to add"(yet again).

PR #830, opened 2018-01-27
Sub render

theKashey:
This implements idea behind #806 - each Component become a AppContainer.

How - just via pre-render hook it will pre-hot-render tree as AppContainer can do.
Difference - it will do it on Component `render`, and force-update afterwards. AppContainer is still needed(preferred) to force sync update.

Solves:
 - async components will reconcile them self. No need to mark components as `hot` - they are all hot
 - solves - a bit of Portals. I am not sure why, but this is not working at 100%. In the example async portal is OK, while sync portal will reset Counter if you will change Counter. And not reset if you will change "hidden counter", which got unmounted in the first case.

Anyway - this feature brings stability, and could solve close to any render quiz, we can get. I am not touching hot-render here, but it could be heavy simplified due to multiple "reconcile" points.


examples/async-portals/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "babel-loader": "^7.1.2",
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-preset-env": "^1.6.1",
+    "babel-preset-react": "^6.24.1",
+    "html-webpack-plugin": "^2.30.1",
+    "webpack": "^3.10.0",
+    "webpack-dev-server": "^2.9.7"
+  },
+  "dependencies": {
+    "emotion": "^8.0.12",
```
neoziro: I think we do not need emotion in this project.

examples/async-portals/src/Counter.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.interval)
+  }
+
+  render() {
+    return (
+      <span>
+        +{this.state.count}:{this.gen++}
```
neoziro: Examples should reflect real examples, I think this is not a real example. I suggest to create a simple Portal example.

neoziro: Or an Async Portal but without tricky things.

examples/async-portals/webpack.config.babel.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = {
+  entry: ['./src/index'],
+  output: {
+    path: path.join(__dirname, 'dist'),
+    filename: 'bundle.js',
+  },
+  module: {
+    rules: [
+      {
+        //exclude: /node_modules|packages/,  // should work without exclude
```
neoziro: Excluding `node_modules` is a good thing, please add it and remove comment.

examples/async-portals/src/async-component.js
```diff
@@ -0,0 +1,37 @@
+import React, { Component } from 'react'
+
+class AsyncComponent extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {}
```
neoziro: We have class properties 😁.

packages/react-hot-loader/src/reconciler/proxyAdapter.js
```diff
@@ -0,0 +1,45 @@
...
+      return {
+        ...element,
+        type: proxy.get(),
+      }
+    }
+  }
+  return element
+}
+
+setStandInOptions({
```
neoziro: We should definitely merge projects.

packages/react-hot-loader/test/hot.dev.test.js
```diff
@@ -63,7 +64,7 @@ describe('hot (dev)', () => {
     wrapper.unmount()
   })
 
-  it('should redraw component on HRM', done => {
+  it.only('should redraw component on HRM', done => {
```
neoziro: nop

examples/async-portals/webpack.config.babel.js
theKashey: As long CRA will not exclude node_modules and some projects might follow the same approach - we shall be ready (and we are ready).

packages/react-hot-loader/test/hot.dev.test.js
theKashey: 😨

examples/async-portals/src/Counter.js
theKashey: I am using them for debug purposes. They reflect only the pure idea of RHL - don't lose the state.
This example is a good one, cos it actually __not__ working for sync portal, as described above.

joshjg:
@theKashey 
>async components will reconcile them self. No need to mark components as hot - they are all hot

Running beta.21, this doesn't seem to be the case. Unless I wrap the async component with `hot(module)` any change causes it to reload completely.

theKashey:
@joshjg - so two questions:
1. how you detect and actually load async component
2. please add this code in any place
```js
 import {setConfig} from 'react-hot-loader';
 setConfig({debug: true});
```
And the check the console output. Something should be inside.

joshjg:
1. Basically the following:
```
componentDidMount() {
    import(
        /* webpackChunkName: "[request]" */
        `modules/${someDynamicValue}`
    ).then(({ default: Component }) => this.setState({ Component }));
```
And webpack handles the rest.
2. I don't see any additional output when using that.

theKashey:
During hot replacement RHL will __not__ trigger `componentDidMount`. If it got executed - so something "upper" remounts.
Please double check, the problem is before async component.
Do you have any HRM on top level?

joshjg:
Correct, I have the root component on the top level wrapped with `hot(module)`. Changing something in the async child causes this root to reload (unless I wrap the async child with `hot(module)` as well).

theKashey:
That's mean that React-Hot-Loader is not working for something between top-level component and loader.
Can you double check, that changing top-level component does not cause the problem, and next moving deeper try to find the "unmounting point".
I am quite interested in it.
>If possible - just share some code.

joshjg:
The loader component is the one that unmounts. None of its ancestors unmount.

theKashey:
So, @joshjg, could you share some code. There must a reason(or a bug) to unmount.

joshjg:
```js
class ModuleLoader extends React.Component {
    static propTypes = {
        moduleId: PropTypes.string.isRequired,
        moduleName: PropTypes.string.isRequired
    };

    state = { Component: null };

    componentDidMount() {
        this.mounted = true;

        import(
            /* webpackChunkName: "[request]" */
            `modules/${this.props.moduleName}/src`
        ).then(({ default: Component }) => {
            if (this.mounted) {
                this.setState(() => ({ Component }));
            }
        });
    }

    componentWillUnmount() {
        this.mounted = false;
    }

    render() {
        const { Component } = this.state;

        if (Component) {
            return <Component {...this.props}/>;
        }

        return <Placeholder/>;
    }
}
```

theKashey:
Normal class 🤷‍♂️ 
Meanwhile I've found a bug which may lead to unmount, and it actually is #843 , but it is not related to sub-render as everyone thought before, and exists from the first beta.
> If you will render one component instead of another - RHL will use the __old__ one.

joshjg:
FYI - beta 22 did not fix this issue

theKashey:
@joshjg, in this case, I'll ask for some example to reproduce.

joshjg:
@theKashey Couple things

- I messed up my babel config - this was causing the unmounts. After fixing this, the loader no longer unmounts, but instead the async child never updates.
- I tried using `setConfig({ logLevel: 'debug' })` (instead of debug: true) and found the following error:
```
React-hot-loader: reconcilation failed due to error TypeError: Cannot read property 'children' of undefined
    at eval (react-hot-loader.development.js:978)
```
Which is this line: https://github.com/gaearon/react-hot-loader/blob/next/src/reconciler/hotReplacementRender.js#L125

theKashey:
I've got the idea, easy to fix, but lets double check.
Some render returned an array as a children (it could be nested), but in render-before thats child did not exist.
```
 { someFlag && [<div>....]
```
Do you have such code?

joshjg:
Yes, I use that pattern all over.

theKashey:
Ok, let me write red tests first.

PR #833, opened 2018-01-29
Compat mode. 

### LINKED ISSUE #775: react-stand-in incorrectly picks up Relay containers as functional components

### LINKED ISSUE #832: `Objects are not valid as a React child` after upgrading `react-stand-in` from `4.0.0-beta.13`
Moved from https://github.com/theKashey/react-stand-in/issues/2

### Description

What you are reporting: Bug

### Expected behavior

What you think should happen: do not throw error

### Actual behavior

What actually happens:

```
Error: Objects are not valid as a React child (found: object with keys {props, context, refs, updater, __reactstandin__proxyGeneration, __reactstandin__cachedResult}). If you meant to render a collection of children, use an array instead.null
    at invariant (invariant.js:42)
    at traverseAllChildrenImpl (react.development.js:805)
    at traverseAllChildren (react.development.js:833)
    at Object.countChildren [as count] (react.development.js:945)
    at doTraverse (index.js:84)
    at doVisit (index.js:116)
    at index.js:172
    at new Promise (<anonymous>)
    at reactTreeWalker (index.js:70)
    at mapper (index.js:88)
```

### Environment

React Hot Loader version: 4.0.0-beta.18
react-stand-in version: >= 4.0.0-beta.14

1. `node -v`: v9.4.0
2. `npm -v`: 5.6.0

Then, specify:

1. Operating system: macOS
2. Browser and version: Chrome 64

### Reproducible Demo

https://github.com/JounQin/react-ssr/tree/react-stand-in

Branch: react-stand-in

For now, I'm locking react-stand-in@4.0.0-beta.13, you can remove resolutions in package.json and run yarn && yarn dev, then visit http://localhost:4000.

theKashey:
fixes #832, breaks #775

The idea - in beta14 we introduced a new behavior for stand-in, to be more clear - returning an instance from stateless components.
This fixes #775, but actually required __only__ by 775/Relay, but in the same time breaks react-async-component and possible related stuff, which can not handle instances from return.

Thus we got a new configuration option - compat, default to true.
In compat mode we will NOT return an instance from stateless components, but just rendering them.
In case of instance - it will throw an error, asking to change configuration.

Instance as result is far rare case, and it possible to detect that you have it, that opposite. That's why it defaults to true.

Not listing a new option in readme, as long one should not configure it before hands.
** Future generations: if you got error "an instance were returned from a render function. Configuration change required" - you know the solution *

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -311,3 +311,19 @@ import * as React from 'react' // React is now patched
 import React from 'react'
 import { hot } from 'react-hot-loader' // React is now patched
 ```
+
+#### React-hot-loader: an instance were returned from a render function. Configuration change required
+
+React-hot-loader runs in "compatibility" mode by default. This disables some workarounds you possible might need.
+
+* **Enable** compat mode if you are using something like `react-async-bootstrapper`(react-async-component)
+* **Disable** compat mode if you are using `Relay`.
+
+There is **no** way to use them simultaneously.
+
+Buy default RHL will produce more simply and compatible code, but non-compact mode is absolutely legit for React 15+.
+
+```js
+import { setConfig } from 'react-hot-loader'
+setConfig({ compat: false })
+```
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -325,5 +325,5 @@ Buy default RHL will produce more simply and compatible code, but non-compact mo
 
 ```js
 import { setConfig } from 'react-hot-loader'
-setConfig({ compat: false })
+setConfig({ statelessIndeterminateComponent: true })
 ```
--- a/examples/styled-components/src/Counter.js
+++ b/examples/styled-components/src/Counter.js
@@ -15,7 +15,7 @@ class Counter extends React.Component {
   }
 
   render() {
-    return this.state.count
+    return <span>1:{this.state.count}</span>
   }
 }
 
--- a/examples/styled-components/src/Counter.js
+++ b/examples/styled-components/src/Counter.js
@@ -15,7 +15,7 @@ class Counter extends React.Component {
   }
 
   render() {
-    return <span>1:{this.state.count}</span>
+    return <span>{this.state.count}</span>
   }
 }
 
--- a/packages/react-hot-loader/src/patch.dev.js
+++ b/packages/react-hot-loader/src/patch.dev.js
@@ -1,6 +1,5 @@
 import React from 'react'
 import reactHotLoader from './reactHotLoader'
-import './utils.dev'
 
 reactHotLoader.patch(React)
 
--- a/packages/react-hot-loader/src/patch.dev.js
+++ b/packages/react-hot-loader/src/patch.dev.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import reactHotLoader from './reactHotLoader'
+import './utils.dev'
 
 reactHotLoader.patch(React)
 
--- a/packages/react-hot-loader/src/reactHotLoader.js
+++ b/packages/react-hot-loader/src/reactHotLoader.js
@@ -74,6 +74,7 @@ const reactHotLoader = {
 
   config: {
     logLevel: 'error',
+    compat: true,
   },
 }
 
--- a/packages/react-hot-loader/src/utils.dev.js
+++ b/packages/react-hot-loader/src/utils.dev.js
@@ -1,9 +1,6 @@
-import { setConfig as setProxyConfig } from 'react-stand-in'
 import { getProxyByType } from './reconciler/proxies'
-import reactHotLoader from './reactHotLoader'
-import logger from './logger'
-
-setProxyConfig({ logger, reactHotLoader: reactHotLoader.config })
+import config from './config'
+import { setConfig as configureStandin } from './reconciler/standInAdapter'
 
 const getProxyOrType = type => {
   const proxy = getProxyByType(type)
@@ -13,4 +10,5 @@ const getProxyOrType = type => {
 export const areComponentsEqual = (a, b) =>
   getProxyOrType(a) === getProxyOrType(b)
 
-export const setConfig = config => Object.assign(reactHotLoader.config, config)
+export const setConfig = newConfig =>
+  configureStandin(Object.assign(config, newConfig))
--- a/packages/react-hot-loader/src/utils.dev.js
+++ b/packages/react-hot-loader/src/utils.dev.js
@@ -1,9 +1,9 @@
+import { setConfig as setProxyConfig } from 'react-stand-in'
 import { getProxyByType } from './reconciler/proxies'
 import reactHotLoader from './reactHotLoader'
 import logger from './logger'
-import { setConfig as setProxyConfig } from 'react-stand-in'
 
-setProxyConfig({ logger })
+setProxyConfig({ logger, reactHotLoader: reactHotLoader.config })
 
 export const areComponentsEqual = (a, b) =>
   getProxyByType(a) === getProxyByType(b)
--- a/packages/react-stand-in/src/createClassProxy.js
+++ b/packages/react-stand-in/src/createClassProxy.js
@@ -87,22 +87,7 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
       result = CurrentComponent.prototype.render.call(this)
     }
 
-    if (isReactComponentInstance(result)) {
-      console.error(
-        'React-hot-loader: ',
-        InitialComponent,
-        'returned an instance',
-        result,
-        'as result.',
-        'You have to disable compat mode to let React-hot-loader handle this.',
-        'setConfig({compat: false}))',
-      )
-      throw new Error(
-        'React-hot-loader: an instance were returned from a render function. Configuration change required',
-      )
-    }
-
-    return wrapResult(result)
+    return wrapResult.call(this, result)
   }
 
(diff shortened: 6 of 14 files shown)


taion:
cc @ctrlplusb 

Is it impossible to patch react-tree-walker to handle this case? It seems like it'd be a pretty odd limitation of react-tree-walker to not work with Relay at all.

theKashey:
https://github.com/ctrlplusb/react-tree-walker/issues/19

theKashey:
Anyway - I expect not only react-tree-walker to fail traverse "the tree". Even we are thinking to simplify traverse, dropping the odd cases and rely on sub-render(#830).

taion:
Async is hard \:

Still, these trees seem traversable. Just another branch to handle for functional components.

docs/Troubleshooting.md
```diff
@@ -311,3 +311,19 @@ import * as React from 'react' // React is now patched
...
+#### React-hot-loader: an instance were returned from a render function. Configuration change required
+
+React-hot-loader runs in "compatibility" mode by default. This disables some workarounds you possible might need.
+
+* **Enable** compat mode if you are using something like `react-async-bootstrapper`(react-async-component)
+* **Disable** compat mode if you are using `Relay`.
+
+There is **no** way to use them simultaneously.
+
+Buy default RHL will produce more simply and compatible code, but non-compact mode is absolutely legit for React 15+.
```
neoziro: This is not very clear, I suggest to change everything and put it in the README :

```
#### Using React Hot Loader + Relay

To enable Relay compatibility mode, you must define a custom config:

__```js
import { setConfig } from 'react-hot-loader'
setConfig({ relayCompatMode: true })
__```

This mode is not compatible with `react-async-bootstrapper` or `react-async-component`, more detail in #833.
```

examples/styled-components/src/Counter.js
```diff
@@ -15,7 +15,7 @@ class Counter extends React.Component {
   }
 
   render() {
-    return this.state.count
+    return <span>1:{this.state.count}</span>
```
neoziro: 1. ?

packages/react-hot-loader/src/patch.dev.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import reactHotLoader from './reactHotLoader'
+import './utils.dev'
```
neoziro: `./utils.dev` should not do anything at runtime, refactoring needed.

packages/react-hot-loader/src/reactHotLoader.js
```diff
@@ -74,6 +74,7 @@ const reactHotLoader = {
 
   config: {
     logLevel: 'error',
+    compat: true,
```
neoziro: Rename to `relayCompatMode` (even if it not just `relay`, for now it is).

packages/react-stand-in/src/createClassProxy.js
```diff
@@ -86,6 +87,21 @@ function createClassProxy(InitialComponent, proxyKey, wrapResult = identity) {
...
+        'React-hot-loader: ',
+        InitialComponent,
+        'returned an instance',
+        result,
+        'as result.',
+        'You have to disable compat mode to let React-hot-loader handle this.',
+        'setConfig({compat: false}))',
+      )
+      throw new Error(
+        'React-hot-loader: an instance were returned from a render function. Configuration change required',
```
neoziro: This is React Stand In, not React Hot Loader.

packages/react-hot-loader/src/utils.dev.js
```diff
@@ -1,9 +1,9 @@
+import { setConfig as setProxyConfig } from 'react-stand-in'
 import { getProxyByType } from './reconciler/proxies'
 import reactHotLoader from './reactHotLoader'
 import logger from './logger'
-import { setConfig as setProxyConfig } from 'react-stand-in'
 
-setProxyConfig({ logger })
+setProxyConfig({ logger, reactHotLoader: reactHotLoader.config })
```
neoziro: If you are doing that, please remove `react-stand-in` and put all code in `react-hot-loader`. I think this is a good idea. We could also remove Lerna.

packages/react-stand-in/src/createClassProxy.js
neoziro: If it is, then we can remove `react-stand-in`.

docs/Troubleshooting.md
theKashey: We do know that "the hack" is needed only for Relay, but it is an absolutely legit structure from React prospective, and someone else could use it. Thus we should not call it `relay` even if we don't know that someone.

neoziro: I agree, but could we find a name more precise than `compat`?

neoziro: Maybe long and explicit, it is OK for me.

theKashey: the "real" name is `supportIndeterminateComponent`, but actually not support, but "use".
`statelessIndeterminateComponent` ?

examples/styled-components/src/Counter.js
theKashey: Eh, just some trash from smoke testing.

packages/react-hot-loader/src/utils.dev.js
theKashey: Let me refactor this a bit. 
Anyway - this is a good decision to make.

theKashey:
This PR will cause merge conflict with sub-render one. It will be a good idea to cover proxy-related code with better tests after merging any of them.

taion:
I'm still not sure this is a reasonable default, though. For example, if you look at npm, react-relay gets ~8x as many downloads a month as react-async-bootstrapper (88k v 11k).

As such, if this is working around a bug in react-async-bootstrapper anyway, then it probably shouldn't be the default behavior.

theKashey:
Both of them actually are quite rarely used, but it is possible to detect relay and ask to change the configuration. We could detect "the instance was returned from a render". 
But we could not detect the opposite situation when that instance will blow up something outside our responsibility. And it impossible to understand that RHL is causing the issue.

taion:
That's pretty reasonable.

I guess the only other factor is that Relay is an official Facebook thing. I dunno. Good point on being able to catch and log on errors w/Relay though.

theKashey:
Maybe it worth invest a bit more time, and make RHL "function" compatible. Ie wrap functions with functions, not classes.

theKashey:
I've spiked "class-less" solution, and here is outcome:
1. Everything works, except `hot-replacement-render`. Stateless components does not leave anything except `type` in traversable react tree. The common issue even for `enzyme` - there is no "instance"
2. I hoped that sub-render PR will mitigate this, and just "continue" hot-render on next class-based component.. but there might be no class-based component in next 10 steps.

Look like this is the only way to solve the puzzle for now.

theKashey:
😿🤘☠️

taion:
What happened?

theKashey:
For now we ago going to support Relay, not React-async-component.

taion:
This seemed like a pretty minimal diff, though. I was quibbling a bit regarding the default behavior, but this seemed like a pretty good and reasonable change.

wKovacs64:
Thanks for all your work on RHL4, it's looking great.

I don't know about Relay or RAC, but I was waiting for this PR to land to fix Enzme shallow rendering of SFC's when using the babel plugin. Maybe I'm just doing something wrong, but this PR fixes it for me.

Simple repro based on RHL's webpack example: https://github.com/wKovacs64/rhl4-enzyme

Am I missing something obvious for this to work? Should I open a new issue? Thanks again.

theKashey:
React-hot-loader should be disabled when you run stuff in jest.
Let me double check your repo.

wKovacs64:
RHL should disable itself under Jest, or am I responsible for disabling it? In my real project, I currently only load the Babel plugin when env is development as a workaround for this issue, but I assumed that was a bit of a hack. Thanks for looking, let me know if I dorked something up!

taion:
RHL doesn't do anything without `module.hot`, which can't possibly be defined in Jest, right?

I guess the alternative is to look for NODE_ENV as `test`?

Given the set of tools that this behavior has the potential to break, though, I guess I'm reluctantly moving toward the view that we Relay users should suffer the pain here 😔

theKashey:
RHL will `completely` disable itself only in production, and `partially` without module.hot. Babel-plugin will be still active, but code it produces - shall not.

wKovacs64:
I guess the Babel plugin being active without module.hot is what's behaving unexpectedly then (differently than RHL3, anyway). Should my demo repo work as-is, or should I disable Babel plugin when NODE_ENV=test?

theKashey:
Our bad. During the last refactoring we just lose that rule about `module.hot`.

wKovacs64:
No sweat, hope you didn't rush on my account. I just thought I just missed an answer in the discussion (I can be a little dense). 😄

theKashey:
The fix was merged into next branch few minutes ago. Will be released soon.

wKovacs64:
Sweet, thanks!

theKashey:
Released among 4.0.0-beta.23

PR #849, opened 2018-02-08
FIX/CI

### LINKED ISSUE #846: Fix CI in v4
Jest is not returning a zero code in v4. The CI seems green but it is red.

brianlmacdonald:
Fixes #846. I removed trailing commas from a getConfig parameter in rollupconfig that was throwing errors in node 6.  

Then I added rollupconfig to prettier ignore so it wouldn't add the commas back in during precommit.

Also added a postinstall build script for .dist


package.json
```diff
@@ -16,6 +16,7 @@
     "lint": "eslint .",
     "prepublishOnly": "yarn build",
     "prebuild": "rm -rf dist",
+    "postinstall": "yarn build",
```
theKashey: As far as I know - postinstall will be executed on a client, without any dev deps, it actually needs to be executed.
Should not work.

brianlmacdonald:
@theKashey I added the bootstraps script and lerna, and removed the postinstall script.

brianlmacdonald:
I wasn't sure of which version to put in the lerna.json, so that may be incorrect.

theKashey:
@brianlmacdonald - no-no-no, we just got rid of lerna.
Please not bring it back, but wipe out a bit more complete as we do.

neoziro:
The only problem to fix is Jest that does not return a zero code. It has nothing to do with Lerna or Rollup. You can just restart from `next` branch, all code here will not help.

neoziro:
Closed in favor of #853

Issue #843, opened 2018-02-05
bug in beta.21


fwh1990:
```
Can only update a mounted or mounting component. This usually means you called setState, replaceState, or forceUpdate on an unmounted component. This is a no-op.
```

webpack 3.10.0
babel 7.0-beta39

Rollback to beta-19 worked.
I don't know why it happened, no more code provide to you, sorry.

neoziro:
This is our number one bug. @theKashey it seems to be introduced by your render PR.

theKashey:
@fwh1990 - so, as long you got a bug - you are the best person to track it down! 🏹 

1. Enable debug mode. setConfig({debug: true}). Maybe it will shed some light (probably not). 
2. Double check that this warning is just a warning. Ie react-hot-loader working correctly.
3. And what is you actually changing. Why something got unmount?

The problem is caused by #806. On each render, and pre-prender call each Component you use will compare some magical number. It will change when you execute a new portion of code (HRM, async-chunk, whatever).
Then, it thinks: hm, maybe it was HRM, lets double check.. and perform `hotReplacementRender`. And next it schedules a force update to all components updated. React will throw a warning if I do it during the render.

The question - why something got rendered, but next got unmounted.

So - what I am asking about - just double check that your application and RHL is working. 
If **yes** - I will just track the mounted state, and don't update freshly unmounted components.
If **no** - so that will a be a problem, as long it should be ok.

skipjack:
@theKashey I'm experiencing this issue too, it's being thrown on a ton of components. I don't fully understand all your points though...

> Enable debug mode. setConfig({debug: true}). Maybe it will shed some light (probably not).

I understand this point based on [this](https://github.com/gaearon/react-hot-loader/tree/next#switch-into-debug-mode).

> Double check that this warning is just a warning. Ie react-hot-loader working correctly.

As far as I could tell the loader still worked fine from a functionality standpoint, however maybe I'm missing something about this question.

> And what is you actually changing. Why something got unmount?

I'm not sure what you mean by this, but...

- One of the components with children in the tree unmounts.
- A varied amount of errors are thrown for components that don't contain any `setState`, `replaceState`, or `forceUpdate` calls.

Hope this helps. Not sure how much else I can dig into this but if needed I might be able to supply a reproducible example. Reverting to `4.0.0-beta.19`, as mentioned by @fwh1990, does fix the issue.

theKashey:
* One of the components with children in the tree unmounts.
[...]

PR #852, opened 2018-02-09
Fix reconciler

theKashey:
This fixes beta.21, which were not broken - the "issue" always were always hidden inside.

Fixes:
- Random parasite warnings (#843). Reason: while rendering `div` hot-render pass thought element from the old instance, "using" the old children of nested components. Thus is uses and updates components which does not exists anymore.
- Parasite warnings from async-loaders (#843). Not register increment generation only on real replacement. Ie no more hot-render after async-load.
- Hot-Reconciler not working if you change child count. Thats is actually still the case - it stops if count is different (PR welcomed), but should work normaly if child is just depends on prop, ie `{ flag && <Component /> }. Beforewards we cleanapp all nulls from the children list, and still doing it, but after the first check.

I've add some tests to cover this cases and manually tested on our examples - all good.


src/proxy/utils.js
```diff
@@ -44,10 +44,11 @@ export function isNativeFunction(fn) {
 }
 
 export const identity = a => a
+const indirectEval = str => [window.eval][0](str)
 
 export const doesSupportClasses = (function() {
   try {
-    eval('class Test {}')
+    indirectEval('class Test {}')
```
theKashey: CI related change. Rollup dislike eval.

fnpen:
Nice! I'm waiting this fix.

Issue #854, opened 2018-02-10
SSR broken in beta.22


neoziro:
Render is now asynchronous so hydrate is broken.

<img width="534" alt="screen shot 2018-02-10 at 21 17 07" src="https://user-images.githubusercontent.com/266302/36066317-dbafa520-0ea7-11e8-93c9-84fe7d0afd09.png">

theKashey:
🤷‍♂️ There is no RHL of server side, there is no RHL side effect before HMR event. How you do it?

theKashey:
@neoziro - still need example to reproduce.

neoziro:
Fixed in beta.23.

PR #857, opened 2018-02-13
No more bugs

theKashey:
Fixes 2 bugs
Improves one case (typeScript)
Adds one case (SSR)

@neoziro, I could not reproduce #854 - is it fine for me.


joshjg:
Fixes the error I was seeing, but the end result (for me) is the same - the async child does not update

PR #859, opened 2018-02-14
"The same prototype"

### LINKED ISSUE #858: react-map-gl breaks when RHL is in place
### Description

Having react-map-gl in a project with RHL, throws errors.

Here's a simple repo
https://github.com/elado/bug-react-hot-loader-react-map-gl

Error is related to bound methods and the way the proxy is applied:

```
Uncaught TypeError: Cannot set property '_mapboxMap' of undefined
    at _mapboxMapLoaded (static-map.js:216)
    at commitAttachRef (react-dom.development.js:8836)
    at commitAllLifeCycles (react-dom.development.js:9951)
    at HTMLUnknownElement.callCallback (react-dom.development.js:542)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:581)
    at invokeGuardedCallback (react-dom.development.js:438)
    at commitRoot (react-dom.development.js:10050)
    at performWorkOnRoot (react-dom.development.js:11017)
    at performWork (react-dom.development.js:10967)
    at requestWork (react-dom.development.js:10878)
    at scheduleWorkImpl (react-dom.development.js:10732)
    at scheduleWork (react-dom.development.js:10689)
    at scheduleTopLevelUpdate (react-dom.development.js:11193)
    at Object.updateContainer (react-dom.development.js:11231)
    at react-dom.development.js:15226
    at Object.unbatchedUpdates (react-dom.development.js:11102)
    at renderSubtreeIntoContainer (react-dom.development.js:15225)
    at Object.render (react-dom.development.js:15290)
    at Object../src/index.js (index.js:7)
    at __webpack_require__ (bootstrap f8f3573aa0535663a8b1:678)
```

This seems to be related to `autobind`
https://github.com/uber/react-map-gl/blob/master/src/components/static-map.js#L79
https://github.com/uber/react-map-gl/blob/b50de38ab0dfed9e976e1d2e6b0c040629bd82b5/src/utils/autobind.js

But in any case, the expectation is that RHL won't break the code.

### Environment

Latest create-react-app - react 16.2, react-hot-loader 4.0.0-beta.22, react-map-gl ^3.2.4

### Reproducible Demo

https://github.com/elado/bug-react-hot-loader-react-map-gl

theKashey:
In other words - transfer original class methods into the Proxy prototype, fixes #858

The idea is to leverage a bit side effect caused by React-Hot-Loader's ProxyComponent, to be more concrete - have the same class members list on the top level prototype. 
I could not say, that this is super common case, but RHL __have__ to be just a bit more _transparent_.

https://github.com/uber/react-map-gl/blob/b50de38ab0dfed9e976e1d2e6b0c040629bd82b5/src/utils/autobind.js
```js
export default function autobind(obj) {
  const proto = Object.getPrototypeOf(obj);
  const propNames = Object.getOwnPropertyNames(proto);

  // propNames contains only ProxyComponent methods (render, mount/unmount)
  // should contain methods from base class

  for (const key of propNames) {
    if (typeof obj[key] === 'function') {
      if (!PREDEFINED.find(name => key === name)) {
        obj[key] = obj[key].bind(obj);
      }
    }
  }
}
```

Related to
https://github.com/uber/react-map-gl/issues/461


neoziro:
@theKashey conflict

theKashey:
resolved, just "squash" merge to hide unnecessary commits.

PR #861, opened 2018-02-16
Add null check

philipnilsson:
I'm encountering an error in `babel.dev.js/exit`. It's in quite a large project, and I don't have time to figure out the reason. (It fails checking a dependency called `js-csp`). Hoping you'll consider adding this extra check.

### DIFF
--- a/src/babel.dev.js
+++ b/src/babel.dev.js
@@ -135,7 +135,7 @@ module.exports = function plugin(args) {
           node[REGISTRATIONS] = null // eslint-disable-line no-param-reassign
 
           // inject the code only if applicable
-          if (registrations.length) {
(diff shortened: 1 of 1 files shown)


theKashey:
The change looks reasonable, and I am going to merge your PR.
The only thing I am asking for - could you change your plugin code to
```js
 exit({ node }, {file}) {
          const registrations = node[REGISTRATIONS]
          if(!registrations) {
              console.error('no program enter', file);
          }
```
next double check the content of that file, and roughly explain (or copy paste code here) how you created a program you can _leave_, but do not _enter_.

I can see the reason to this in the current implementation, it might be some serious bug, hiding between lines.

PR #862, opened 2018-02-18
Update README.md

blade254353074:
The sentence should be '**React Hot Loader** is not compatible with these two webpack plugins'


theKashey:
Actually, the problem is with ExtractTextPlugin, and only it has to be deactivated.

blade254353074:
Oh...I misunderstood. Thanks ~ @theKashey

PR #863, opened 2018-02-18
Disable RHL in production, and when HMR not enabled

theKashey:


neoziro:
@theKashey I guess this is what broke SSR right?

theKashey:
May be. The SSR example, I’ve written was ok even without patch.

Issue #866, opened 2018-02-19
react-stand-in is missing


iamstarkov:
### Description

there are 2 links to `react-stand-in` in the ["Known limitations and side effects" of the README](https://github.com/gaearon/react-hot-loader/tree/v4.0.0-rc.0#known-limitations-and-side-effects). There is nothing there if you follow the links https://github.com/gaearon/react-hot-loader/tree/next/packages/react-stand-in

### Expected behavior

Something should be there

### Actual behavior

nothing is

### Environment

v4.0.0-rc.0 https://github.com/gaearon/react-hot-loader/tree/v4.0.0-rc.0

theKashey:
Somewhere between beta.19 and beta.21 stand-in was merged into main project.
Here it is - https://github.com/gaearon/react-hot-loader/tree/next/src/proxy

TODO: fix links.

PR #868, opened 2018-02-20
Fix links

moondef:
#866

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,14 +401,14 @@ export default hot(module)(App)
 
 On code replace you are replacing the old code by a new one. You should not use
 the old code, as thus allow RHL to safely modify it. See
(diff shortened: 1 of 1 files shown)


theKashey:
I am not sure that this is enough. Could you please drop the [old readme](https://github.com/gaearon/react-hot-loader/blob/v4.0.0-beta.19/packages/react-stand-in/README.md) for stand-in, the old link actually refers to, in this directory?

moondef:
@theKashey done

theKashey:
I mean - physically drop(restore) the file in the correct place. 
And you should not refer to the "obsolete branch", or even "current branch name". Sooner or later _next_ will become _master_, and what then?

neoziro:
We should not refer stand-in anymore, please replace next by master and point to an existing file.

PR #870, opened 2018-02-21
Fix IE 11 issue

oliviertassinari:
People might not have a polyfill for `includes()`.
https://github.com/gaearon/react-proxy/blob/13f076b17b43a9d53c151931f3629ef1baae42e5/src/createClassProxy.js#L71

capture d ecran 2018-02-21 a 20 40 39


theKashey:
Good catch!

neoziro:
Thanks @oliviertassinari

Issue #912, opened 2018-03-28
Random unmounts with code splitting


theKashey:
### The code
```js
const Component1 = () => {}
const Component2 = () => <Component1 />

export default Component1
export {
  Component2
}
```
Component1 and default export has the same types, but, was registered by different names twice.
```js
import Component3 from './Component1';
export default Component3;
```
Component3 is still the same, but yet again has a "named" Proxy.

### The problem
You got 3 "same" components, which were the same only in the beginning, but now all are unique ProxyFacades.

React hot loader will call `getProxyByType` passing the type to get the proxy. And only one `name` is associated by type - __the last one__.

During the code splitting one could touch a new code and move "the last one" to a new "name", thus a new ProxyFacade.

Just next loading of async chunk will trigger sub render to reconcile trees, it will found "a very similar components", and update existing components... resulting unmount, as long component type did change.

### The fix
Only one ProxyFacade should be created for any Component. Also doable if we will not `.register` variables at all, but it will break loadable-components HMR (@neoziro).

I am thinking about extracting [type comparison](https://github.com/gaearon/react-hot-loader/blob/master/src/reconciler/hotReplacementRender.js#L294) from reconsiler and using it inside .register

theKashey:
We already "fixed" it - https://github.com/gaearon/react-hot-loader/blob/master/src/proxy/createClassProxy.js#L62
But we do break the fix.

PR #915, opened 2018-03-30
Double proxy registration. Fix regression #912

theKashey:
#912 - Super nasty bug :(, @leandrorlemos - kudos for reporting.

Could break all the things in case of code splitting - something "old" could be "registered" by a new name (it is just a variable name), obtain a new proxy and unmount old instances.

Was fixed in v3. We broke it during migration, and had no tests around it.

### DIFF
--- a/src/proxy/createClassProxy.js
+++ b/src/proxy/createClassProxy.js
@@ -19,7 +19,11 @@ import { inject, checkLifeCycleMethods, mergeComponents } from './inject'
 
 const has = Object.prototype.hasOwnProperty
 
-const proxies = new WeakMap()
+let proxies = new WeakMap()
+
+export const resetClassProxies = () => {
+  proxies = new WeakMap()
+}
 
 const blackListedClassMembers = [
   'constructor',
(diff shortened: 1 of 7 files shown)


src/proxy/createClassProxy.js
```diff
@@ -309,7 +315,9 @@ function createClassProxy(InitialComponent, proxyKey, options) {
 
   update(InitialComponent)
 
-  const proxy = { get, update }
+  proxy = { get, update }
+
+  proxies.set(InitialComponent, proxy)
```
theKashey: This is actually a fix. Everything else is related to this change - now tests have "reset" proxies sometimes.

test/AppContainer.dev.test.js
```diff
@@ -1202,6 +1202,55 @@ describe(`AppContainer (dev)`, () => {
...
+      <AppContainer>
+        <App />
+      </AppContainer>,
+    )
+    expect(spy).not.toHaveBeenCalled()
+    wrapper.setProps({ children: <App /> })
+    expect(spy).not.toHaveBeenCalled()
+    RHL.register(App, 'App3', 'test.js')
+    wrapper.setProps({ children: <App /> })
+    expect(spy).not.toHaveBeenCalled()
```
theKashey: It will stop here.

PR #916, opened 2018-03-30
Move opencollective to dev deps

theKashey:
```text
$ opencollective postinstall

     *** Thank you for using react-hot-loader! ***

Please consider donating to our open collective
     to help us maintain this package.

  https://opencollective.com/react-hot-loader/donate

                    ***
```

💩💩🤷‍♂️

It will not help to find backers. I think this is doing something opposite.

By moving to dev deps we still able to display backers and sponsors in Readme, but will not bother end users and CI tools

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "husky": "^0.14.3",
     "jest": "^22.3.0",
     "lint-staged": "^6.1.1",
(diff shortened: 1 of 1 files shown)


neoziro:
I think it is too much yes, I will remove it completely.

Issue #873, opened 2018-02-23
Don't wrap SFC with Stateful ProxyComponent


theKashey:
Currently, we wrap all the functional component into the stateful ones, and this is not very cool.
- Why do we need it? To have an `instance` to be used in `hot-render`
- Why do we need that instance? Cos we are lazy, and want just `component.render()`
- Can we stop doing it?

Actually yes - hot-replacement-render still have a code for SCF, the only part it is missing - `context`. As long React 15 has context as a part of react-tree, React 16 has not.

It is possible to render Components as we did it all the time, and tract the `context` between two Components as `react-tree-walker` is doing it (https://github.com/ctrlplusb/react-tree-walker/blob/master/src/index.js#L171) - just _Object.assign({}, context, instance.getChildContext())_

__The problem__ - sub-render. One could not start rendering in the "middle", as long there is no way to get `whole` context state in a random position.
I hope some dirty hacks could provide us desired global context variable( ie `this._renderer._context`), but I don't know them yet.

theKashey:
Ok. If some `function` does not has `.contextTypes` - that mean it does not has context. As result - we could continue wrapping only _context-consuming_ SFC, keeping the _normal ones_ unchanged.

This will make most of functional components.. functional again!

theKashey:
I have spiked possible solution, and everything should work IF that everything had Component in the beginning.
Things become complicated if not, as long there is no way to `forceUpdate` SFC. There is no reaction even if one will `enqueueForceUpdate` manually.

In the same time - "just updated" always work, as long replace one function call by another - it's an easy task.

theKashey:
There is an issue with implementation. Not all props got merged correctly.

theKashey:
Fixed

PR #919, opened 2018-04-01
Don't wrap SFC by Stateful ones

theKashey:
Source issue : #873

Don't wrap components we could not wrap. And we could not wrap any component without context types.
In reality - this is almost all SFC.

Side effect: one should have [Stateful] Component before SFC, or it will not be updated on hot-replace.

All tests are green, but there is one little moment bothering me a bit...


test/reconciler.test.js
```diff
@@ -88,6 +92,7 @@ describe('reconciler', () => {
           <ComponentSwap>42</ComponentSwap>
         </root.Component>
       )
+      App.contextTypes = {}
```
theKashey: Force Statefull one to have update during the tests.

test/reconciler.test.js
```diff
@@ -218,8 +231,8 @@ describe('reconciler', () => {
 
       incrementGeneration()
       wrapper.setProps({ second: false })
-      expect(First.rendered).toHaveBeenCalledTimes(5)
-      expect(Second.rendered).toHaveBeenCalledTimes(4)
+      expect(First.rendered).toHaveBeenCalledTimes(6)
```
theKashey: This is "the main concern" -  I have no idea why this is happing, no explanation, nothing...

theKashey:
Checked, double checked, and ready for review. I am not sure why codeconv complain.

Issue #918, opened 2018-03-31
Support React 16.3


theKashey:
Currently, we rely on `componentWillReceiveProps` to perform sync update and sync `forceUpdate`. So - this method does not exist anymore.

We have 2 options:
 1 - rely on subrender, ie async updates. It is a dev mode, nobody cares about double rendering on rare hot updates. Thus remove AppContainer completely. This will also solve 
 2 - find another way. But I don't see any.

oliviertassinari:
Would it also explain why I see a warning anytime is use `getDerivedStateFromProps()`? I'm assuming react-hot-loader is hijacking the components and add `componentWillReceiveProps `.
![capture d ecran 2018-03-31 a 13 20 33](https://user-images.githubusercontent.com/3165635/38162706-8606e736-34e6-11e8-8b7b-3b48bbccaba8.png)

theKashey:
++ we have to test how new ("real") Context API works, as long "Context" is not a "Tag" or a function - it is a brand new element type. Should be fine, but even enzyme itself failing :)

theKashey:
@oliviertassinari - yeah, RHL is adding componentWillReceiveProps, and I am going just to drop it.

swashata:
Had just this issue right now. I am very new to react, just started and had to scratch my head (like a lot) to understand this isn't coming from mine, rather the hot loader 😁 . I will be waiting for an update.

gaearon:
I don't understand what this means:

>rely on subrender, ie async updates.

theKashey:
@gaearon - this is a local meme( #830 ), about performing _hot-render_ each time you "can do it". V3 uses AppContainer to start an update, v4 can use AppContainer to perform sync updates, and Component.render to perform async updates. Async cos they could be executed after the main update or main update could not "render into them" - code splitting, resource load, portals, and so on. Plus we have to defer .forceUpdate, as long shall not call it from a render method.
Basically - this one line - https://github.com/gaearon/react-hot-loader/blob/master/src/proxy/createClassProxy.js#L164

gaearon:
What are you using `componentWillReceiveProps` for today?

theKashey:
Nothing but to being able "flush"(forceUpdate) changes. As I pointed above - we could safely remove it. It may cause double renders, but this is not something we should avoid.

gaearon:
Got it. `componentWillReceiveProps` is a bad place to call `forceUpdate` so that sounds good.

[...]

Issue #860, opened 2018-02-14
[v4] StrictMode warns even when no unsafe methods are used


apostolos:
### Environment

React Hot Loader version: 4.0.0-beta.22
React: 16.3+

### Description

When enabling Strict/Async mode for any part of the tree that contains React components, React warns about unsafe lifecycle methods even when the components don't use them.

This only happens when *react-hot-loader/babel* plugin is loaded.

### Expected behavior

There should be no warnings.

### Actual behavior

```
Warning: Unsafe lifecycle methods were found within a strict-mode tree:

componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Test

componentWillUpdate: Please update the following components to use componentDidUpdate instead: Test

Learn more about this warning here:
https://fb.me/react-strict-mode-warnings
```

### Reproducible Demo

```jsx
import * as React from 'react';
import { render } from 'react-dom';

class Test extends React.Component {
  render() {
    return <div>test strict</div>;
  }
}

render(
  <React.StrictMode>
    <Test />
  </React.StrictMode>,
  document.getElementById('app')
);
```

theKashey:
- `componentWillUpdate` is already removed
- `componentWillMount` to be replaced by `componentDidMount`
- we could not replace `componentWillReceiveProps` yet, but as long it will coexist with React until version 17 - we might just prefix it by "UNSAFE" for a while.

JustFly1984:
Sorry to bug you out, but I have issues with `<StictMode />`  in `gatsby v2`. it has `react-hot-loader@4.3.12` in dependencies, and I'm using it for example in my attempt to make new google maps api for React. I need to test it with SSR, and at the same time to test it in strict mode. It seems like I can't do both at the same time.

JustFly1984:
Should I make new issue? I've made an issue in gatsby: https://github.com/gatsbyjs/gatsby/issues/10179

PR #927, opened 2018-04-06
Making RHL more React 16.3 compatible #918

theKashey:
removing componentWillReceiveProps #918, #860

"sync" flush moved to componentDidUpdate.
"render" also removes scheduled update, cos we are not "double rendering" updated components as we did before.
removing componentWillReceiveProps from AppContainer might change the way error overlay is shown, but we do not actually support that overlays in version 4.

TODO:
- [x] -  add react-lifecycle-compact
- [x] -  double manually test

### DIFF
--- a/examples/styled-components/package.json
+++ b/examples/styled-components/package.json
@@ -20,7 +20,7 @@
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "react-emotion": "^8.0.12",
-    "react-hot-loader": "next",
+    "react-hot-loader": "^4.0.1",
     "styled-components": "^2.4.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "global": "^4.3.0",
     "hoist-non-react-statics": "^2.5.0",
     "prop-types": "^15.6.1",
+    "react-lifecycles-compat": "^2.0.0",
     "shallowequal": "^1.0.2"
   },
   "engines": {
--- a/src/AppContainer.dev.js
+++ b/src/AppContainer.dev.js
@@ -1,20 +1,10 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { polyfill } from 'react-lifecycles-compat'
 import logger from './logger'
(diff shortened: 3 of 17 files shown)


neoziro:
Thanks for taking care of that!

Fer0x:
Steps to use this PR (for webpack users):

1. Add dependency with `"react-hot-loader": "gaearon/react-hot-loader#927/head"` in `package.json`.
2. Add to `scripts` directory in `package.json` following line:
```    "postinstall": "(cd node_modules/react-hot-loader && yarn && yarn build)"```
3. Finally, in your webpack config, add this to `resolve.alias` section:
`react: './node_modules/react'`

colinrcummings:
Thanks, @Fer0x. Will these steps still be required once this PR lands in `master`?

theKashey:
@colinrcummings - sure not.
I'll try to handle this one today. 🤞

theKashey:
🕷 do not merge. It's killing React 15.

Fer0x:
@theKasher what about release `5.x` major version? `4.x` will remain for React 15

theKashey:
Should be ok by now. `instanceof React.Component` is not properly working to detect SFC, _and I have no idea why_.
As result ProxyComponent tracks it state by itself via static property. That property sometimes, on component update got copied from function to classes, _and I have no idea why_.

I've added "stack trace" to hot-render, to help myself debug this odd behavior. As result "the fix" - is just to add `isStatelessFunctionalProxy` to the RESERVED_KEYWORDS and stop coping it.
This fixes the issue, even if I did not found _when and why_ my stateful components were soiled. So - __something bad could still exists__. But tests are green, and most complex application I've tried - also works well.

As result I keen stack tracing, as long it is __still__ useful, as long RHL actually could not "merge children" from time to time. And, yet again, _and I have no idea why_.

Folks, you have few hours to test last version in React 15 (nothing was changed for React 16) and next I merge.

theKashey:
There is SFC component related but, introduced by https://github.com/gaearon/react-hot-loader/issues/873 :(
I will try to fix it, or have disable this feature (it is not in production yet)

Issue #947, opened 2018-04-24
react-hot-loader not work


MrZhang123:
### Description

I use react-hot-loader and webpack for HMR, i use `hot` for App , but when i save the code , hot reload not work , how i change ?

### Environment

node: 8.9.0
react-hot-loader:  4.1.1
webpack :4.6.0
webpack-dev-server: 3.1.3

### Code

webpack.config.js

```JS
module.exports = {
  entry: {
    main: [
      'babel-polyfill',
      path.resolve(__dirname, '../src/main.js')
    ]
  },
  output: {
    path: path.resolve(__dirname, '../dist'),
    publicPath: './',
    filename: '[name].js',
    chunkFilename: 'chunk/[name].[chunkhash].js'
  }
}

// --------

config.entry.main = (config.entry.main || []).concat([
  'react-hot-loader/patch',
  `webpack-dev-server/client?http://localhost:${PORT}/`,
  'webpack/hot/dev-server'
])
config.plugins = (config.plugins || []).concat([
  new webpack.HotModuleReplacementPlugin()
])
config.mode = 'development'

const compiler = webpack(config)

const server = new WebpackDevServer(compiler, {
  hot: true,
  noInfo: true,
  quiet: true,
  filename: config.output.filename,
  publicPath: '/',
  stats: {
    colors: true
  }
})
```

App.js

```JS
import { hot } from 'react-hot-loader'

import RouterLink from './views/RouterLink'

class App extends Component {
  render () {
    return <RouterLink />
  }
}

export default hot(module)(App)
```

store.js

```JS
const configureStore = () => {
  const store = createStore(
    reducers,
    applyMiddleware(thunk, middleware)
  )

  if (process.env.NODE_ENV !== 'production'){
    if (module.hot) {
      module.hot.accept('../reducers', () => {
        store.replaceReducer(reducers)
      })
    }
  }
  return store
}
```

main.js

```JS
import configureStore from './store'

const history = createHistory()
export const store = configureStore()

ReactDOM.render((
  <Provider store={store}>
    <ConnectedRouter history={history}>
      <App />
    </ConnectedRouter>
  </Provider>
), document.getElementById('app'))
```

theKashey:
Please double check console output. It should contain something.
First thing to check - does webpack HMR working, as long it might not.
Usually, then RHL does not work - you will see the change, after page reload. But if HMR does not work - you will see _nothing_.

theKashey:
[...]

PR #963, opened 2018-05-03
doc: more clear statement about code splitting

theKashey:
Related to https://github.com/gaearon/react-hot-loader/issues/947
A bit more clear statement about code splitting..

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,24 +203,35 @@ Using React Hot Loader with React Native can cause unexpected issues (see #824) 
 
 ### Code Splitting
 
-Most of modern React component-loader libraries ([loadable-components](https://github.com/smooth-code/loadable-components/),
-[react-loadable](https://github.com/thejameskyle/react-loadable)...) are compatible with React Hot Loader.
+Most of modern React component-loader libraries are not "100%" compatible with React-Hot-Loader, as long
+they load deferred component on `componentWillMount`, but never reload it again.
+As result - you can update the code, webpack will ship that code the the client, but nothing will be updated.
 
-You have to mark your "loaded components" as _hot-exported_.
+In this case, you have to setup "reloading" by your self - you have to mark your "loaded components" as _hot-exported_.
 
-Example using
-[loadable-components](https://github.com/smooth-code/loadable-components/):
+Example using [react-async-component](https://github.com/ctrlplusb/react-async-component)...)
 
 ```js
 // AsyncHello.js
-import loadable from 'loadable-components'
-const AsyncHello = loadable(() => import('./Hello.js'))
+import asyncComponent from 'react-async-component'
+const AsyncHello = asyncComponent({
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -203,25 +203,36 @@ Using React Hot Loader with React Native can cause unexpected issues (see #824)
...
+import asyncComponent from 'react-async-component'
+const AsyncHello = asyncComponent({
+  resolve: () => import('./Hello.js'),
+})
 
 // Hello.js
 import { hot } from 'react-hot-loader'
 const Hello = () => 'Hello'
-export default hot(module)(Hello) // <-- the only change to do
+export default hot(module)(Hello) // <-- module will reload itself
```
neoziro: You should use `loadable-components` or `react-imported-component`. They provide the best support for RHL.

README.md
```diff
@@ -203,25 +203,36 @@ Using React Hot Loader with React Native can cause unexpected issues (see #824)
 
 ### Code Splitting
 
-Most of modern React component-loader libraries ([loadable-components](https://github.com/smooth-code/loadable-components/),
-[react-loadable](https://github.com/thejameskyle/react-loadable)...) are compatible with React Hot Loader.
+Most of modern React component-loader libraries are not "100%" compatible with React-Hot-Loader, as long
```
neoziro: We should discourage using `react-loadable`.

neoziro:
I made some test on Loadable Components. I don't know what has changed but I no longer need to reload everything. It works out of the box without adding any code. Could be the same for React Loadable.

theKashey:
You know, it sounds like a bug. Ok a huge change inside webpack. 
It could only work is something triggered “register”. 
Can you share that example?

neoziro:
Example from [this PR](https://github.com/smooth-code/loadable-components/pull/73). Let's make it clear before rewriting this README section.

theKashey:
@neoziro - so everything is due to autoload inside the constructor. Look like the condition should be a bit different, - preload only NOT for browser, to emulate `componentWillMount` on server-side.
In [imported-component](https://github.com/theKashey/react-imported-component/blob/master/src/Component.js#L36) I am using `detect-node` for it. Fix and test this moment just last weekends, so I am pretty sure you shall do the same.

As result - the state of code splitting is unchanged. You still have somehow trigger reimport, and it will not work without it.

neoziro:
@theKashey the behaviour that I want is loading in component but only server-side https://github.com/smooth-code/loadable-components/blob/master/src/loadable.js#L58. So I think doing it in constructor makes it compatible with HMR, so nice!

Yes the state of Code Splitting stay unchanged. Loadable Components and Imported Components work both out of the box.

theKashey:
But `typeof window !== 'undefined'` is a browser, not server side.
Anyway, this one is good to merge.

PR #967, opened 2018-05-08
Props shouldn’t be overwritten

timneutkens:
I'm not sure what the correct functionality is, either providing the new props vs the old ones / merging them etc. But the change made in 4.1.0 is breaking and should be part of a major version.

It also seems weird to me that `componentWillUpdate` is called from react-hot-loader 🤔

Fixes https://github.com/zeit/next.js/issues/4299

screen shot 2018-05-08 at 11 21 00

`testProps` is the old way they were calculated in 4.0.0/4.1.0

`nextProps` are the new way they were calculated in 4.1.0+

My change brings back the result of `testProps` rather than `nextProps`.

### DIFF
--- a/src/reconciler/hotReplacementRender.js
+++ b/src/reconciler/hotReplacementRender.js
@@ -262,8 +262,8 @@ const hotReplacementRender = (instance, stack) => {
       // copy over props as long new component may be hidden inside them
       // child does not have all props, as long some of them can be calculated on componentMount.
       const nextProps = {
-        ...instance.props,
         ...(child.nextProps || {}),
+        ...instance.props,
         ...(child.props || {}),
       }
 
--- a/test/reconciler.test.js
+++ b/test/reconciler.test.js
@@ -128,7 +128,7 @@ describe('reconciler', () => {
 
       // what props should be used? Look like the new ones
       expect(second.willUpdate.mock.calls[0]).toEqual([
-        { children: '42', newProp: true, keyId: '2' },
+        { children: '42', newProp: true, keyId: '1' },
         null,
         { children: '42', keyId: '1' },
         null,


timneutkens:
Another solution to the problem we're having with styles-jsx is remove the extra `componentWillUpdate`

theKashey:
> It also seems weird to me that componentWillUpdate is called from react-hot-loader 🤔
This is how React-Redux got signal to update wrapped component.

As far as I know - the logic around merging props should correct.
```js
const Switch  = ({flag}) => flag ? <button1 /> : <button2 />;

<Switch flag={0}
// and then
<Switch flag={1}
```
React will re-render something with flag={1} and that means that we have to work with flag={1}, with a new prop, not the old one.

Look like, you dont need this change. You might need remove side effect on props. So - store the original props, and restore them just after render.

> in the same time, I am merging old tree and a new tree. Probably I should use the values used to build existing tree, ie "the old values".

I have to double check.

Meanwhile - could you double check that
```js
const next = instance => {
      // copy over props as long new component may be hidden inside them
      // child does not have all props, as long some of them can be calculated on componentMount.
      const realProps =  instance.props; // <<-------
      const nextProps = {
        ...instance.props,
        ...(child.nextProps || {}),
        ...(child.props || {}),
      }

      if (isReactClass(instance) && instance.componentWillUpdate) {
        // Force-refresh component (bypass redux renderedComponent)
        instance.componentWillUpdate(nextProps, instance.state)
      }
      instance.props = nextProps
      hotReplacementRender(instance, stackChild)
      instance.props = realProps; // <<-----
    }
```
Fixes your problem. Might be this change should be added.

timneutkens:
> Meanwhile - could you double check that

Since:

```
if (isReactClass(instance) && instance.componentWillUpdate) {
        // Force-refresh component (bypass redux renderedComponent)
        instance.componentWillUpdate(nextProps, instance.state)
      }
```

is still called the result is the same

timneutkens:
The code is located here: https://github.com/zeit/styled-jsx/blob/master/src/style.js#L26-L28

theKashey:
Why changing the props merge order (and why you are getting new values?) fixes the problem?

theKashey:
Ok. So what I found:
1. The merge order was right. There is no difference between `child.props` and `nextProps`, they are both is `props` of Component, just from different sources (lets call it so)
2. That might be not _correct_ in real, but this is how v4 works and this is how we build out tests.
3. Your problem is not due to props, but due to how `componentWillUpdate` called. As long _"we are not touching lifecycle events"_, probably we should call this method with unchanged props.
4. There were a side effect with soiling component props, I noticed with React 15. The patch will fix it.

See https://github.com/gaearon/react-hot-loader/pull/968

timneutkens:
Awesome @theKashey 🏆 ❤️ 

> probably we should call this method with unchanged props

Yeah so this is what was done before (in rhl 3 / 4.0.0/4.0.1), I'll check your branch 👍

timneutkens:
#968 fixes this issue.

timneutkens:
Will be fixed when https://github.com/gaearon/react-hot-loader/pull/968 is merged.

PR #970, opened 2018-05-10
fix: construction sideEffect

theKashey:
This PR removes "construction" side effect, @neoziro found while debugging Loadable-components.

In short - RHL will construct Components to examine internal changes, even if user never used them in real.

This PR making this operation possible only for mounted components, postponing until usage for the rest.


theKashey:
Sweet. Do not release until `this` related issue will be solved.

theKashey:
@neoziro  - today?

Issue #969, opened 2018-05-09
Uncaught ReferenceError: _this5 is not defined


timneutkens:
If you are reporting a bug or having an issue setting up React Hot Loader, please fill in below. For feature requests, feel free to remove this template entirely.

### Description

After hot-reloading a file with multiple components (4 to 5) in one file an error is thrown

```
VM97:6 Uncaught ReferenceError: _this5 is not defined
    at value (eval at __reactstandin__regenerateByEval (header.js:996), <anonymous>:6:9)
    at HTMLUnknownElement.callCallback (react-dom.development.js:540)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:579)
    at Object.invokeGuardedCallback (react-dom.development.js:436)
    at Object.invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:450)
    at executeDispatch (react-dom.development.js:834)
    at executeDispatchesInOrder (react-dom.development.js:856)
    at executeDispatchesAndRelease (react-dom.development.js:954)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:965)
    at Array.forEach (<anonymous>)
```

### Expected behavior

Hot reload should be applied and not crash the component

### Actual behavior

What actually happens:

`_this5 is not defined`

timneutkens:
Sounds very similar to #928

theKashey:
There was `_this`, `this2`, `this3`, half year later someone asked for `this4`, and now `this5`.
😩🤘

loganfsmyth:
I'm happy to try to help, but it would go a long way if someone could spell out what this wrapper is doing, since I'm not familiar with the internals of the hot loader.

What ends up running at the end, and why isn't `_this5` already available in the outer scope?

Can could in theory generate any number of `_this` variables, dependent only on how many arrow functions and `this` references you have in a single file.

theKashey:
So what we are doing - reading function body from one class, and executing it in the scope of another. as result all variables, possible _visible_ from a function, including `this`.

@loganfsmyth - this problem could be fixed with easy, if only you know babel. And look like - you know.

We just need to place our "regenerator" function inside a constructor, the place where all those `thisN` already defined. Currently, we are just adding it to the class.
The trick is - we have to the function into the transpiled class constructor, so out of the scope, where this5 was defined.
Link to a sources - https://github.com/gaearon/react-hot-loader/blob/master/src/babel.dev.js#L176

[...]

PR #972, opened 2018-05-12
need moar this

theKashey:
Just adding more this for #969


theKashey:
@timneutkens - could you try master branch? Is this the last fix we all need?

timneutkens:
Solves the issue we had.

theKashey:
Ok, lets wait for #975 and then release a new version.

Kaishiyoku:
Same issue here with this5 - when will this fix be released?

theKashey:
@Kaishiyoku - hold my beer

theKashey:
version 4.2.0 - released.

PR #973, opened 2018-05-12
docs(readme): simplify code splitting

neoziro:


neoziro:
I have nothing against Jamie approach, it is just important to notice it. I remove the <s>striked line</s> if you want.

theKashey:
I would appreciate. Closed issues is not a technical problem.

PR #981, opened 2018-05-17
docs(readme): Better type comparison

theKashey:


README.md
```diff
@@ -267,6 +267,26 @@ const element = <Component />
...
+Another way - compare "rendered" element type
+
+```js
+const element = <Component />
+console.log(element.type === (<Component/>).type) // true
+
+// better - precache rendered type
+const element = <Component />
+const ComponentType = (<Component />).type
+console.log(element.type === ComponentType // true
```
neoziro: Miss parenthesis

README.md
```diff
@@ -267,6 +267,26 @@ const element = <Component />
...
 ```
 
+Another way - compare "rendered" element type
+
+```js
+const element = <Component />
+console.log(element.type === (<Component/>).type) // true
+
+// better - precache rendered type
+const element = <Component />
```
neoziro: Incompatible with PropTypes

neoziro:
I think this is not a good thing to add in README, this is very confusing and it will generate a lot of issues.

README.md
theKashey: :( Yeah, now I recall all the issues around it.

PR #982, opened 2018-05-17
docs(readme): Better type comparison

theKashey:


theKashey:
Actually - there is a non RHL specific way to compare types (but a bit 💩)
```js
element.type = React.children.only({type: Component}).type
```

I am just looking for anyway, we could provide for a library creators. The ones who could not use any methods RHL could provide.

PR #994, opened 2018-05-28
Fixes #913 - Fix/improve guide for Electron

### LINKED ISSUE #913: Fix/improve guide for Electron
HMR doesn't seem to work as expected in that example that is [linked to in the README](https://github.com/gaearon/react-hot-loader#electron). 

If I merely extend it to have the following component structure:
```
App
|-- Branch1
|-- Branch2
```
... and change something in Branch2 ...

### Expected behavior

... I expect only Branch2 and App to be updated.

### Actual behavior

... instead, this happens:
```
HMR sent to all windows!
  ←[3crimson;1melectron-compile:require-hook ←[0mGot HMR signal! ←[3crimsonm+41s←[0m
  ←[3crimson;1melectron-compile:require-hook ←[0mRemoving node module entry for D:\development\github\hmr-example-issue-2\src\app.jsx ←[3crimsonm+3ms←[0m
  ←[3crimson;1melectron-compile:require-hook ←[0mRemoving node module entry for D:\development\github\hmr-example-issue-2\src\branch1.jsx ←[3crimsonm+1ms←[0m
  ←[3crimson;1melectron-compile:require-hook ←[0mRemoving node module entry for D:\development\github\hmr-example-issue-2\src\branch2.jsx ←[3crimsonm+0ms←[0m
```

### Environment

React Hot Loader version: 4.0.0

### Reproducible Demo

https://github.com/s-h-a-d-o-w/hmr-example-issue-2

### Further notes

I think it should be mentioned in the README that the guide only applies to electron-forge/electron-compile. There is no `.compilerc` if you use e.g. electron-webpack or some custom electron setup.

On a sidenote - the behavior is as expected when using electron-webpack and the RHL v3 workflow.

But if you do stick with electron-compile, I think the code snippet in the README should be a bit bigger, e.g.:
```
import {enableLiveReload} from 'electron-compile';

const isDevMode = process.execPath.match(/[\\/]electron/);
if(isDevMode) {
    enableLiveReload({ strategy: 'react-hmr' });
}
```

s-h-a-d-o-w:
Apologies for the delay. I wanted to see how the situation with electron-compile's HMR would develop before committing to a recommendation.

And I had to come up with a good workflow myself. Because it was only after I submitted issue #913 that I had the fairly banal realization that one could simply view an Electron app like any other web app. And that maybe one should. Most people probably choose to use Electron because they are already familiar with JS. So they've probably already used webpack. Why introduce new tools that at least partly don't work as well as webpack does?

But I left the reference to electron-compile in there because it is quite popular, it does require less configuration and some people have argued that webpack should not be used with Electron. (Although they possibly overlook the fact that parse time may still play a role even when a big project is loaded directly from one's hard drive...)


README.md
```diff
@@ -183,22 +183,11 @@ We also have a [full example running Parcel + React Hot Loader](https://github.c
...
-  "plugins": ["react-hot-loader/babel"]
-}
-```
-
-2.  Enable Live Reload in the project
-
-```js
-enableLiveReload({ strategy: 'react-hmr' })
-```
+One way of doing this with Electron is to simply use webpack like any web-based project might do and the general guide above describes. See also [this example Electron app](https://github.com/s-h-a-d-o-w/rhl-electron-quick-start)
```
neoziro: Webpack*

README.md
```diff
@@ -183,22 +183,11 @@ We also have a [full example running Parcel + React Hot Loader](https://github.c
...
 
-See a [complete example](https://github.com/rllola/hmr-example-issue-2/blob/master/src/index.js).
+A webpack-less way of doing it to use `electron-compile` (which is also used by `electron-forge`) - see [this example](https://github.com/rllola/hmr-example-issue-2). While it requires less configuration, something to keep in mind is that `electron-compile`'s HMR will always reload all modules, regardless of what was actually edited.
```
neoziro: Webpack-less*

- Links on electron-compile and electron-forge

s-h-a-d-o-w:
Hm... I didn't see that and you're obviously right for consistency's sake.

However, I would like to point out that the official way of spelling "webpack" is indeed with a lower-case w in the beginning: https://webpack.js.org/concepts/
So it seems to me that instead, all occurrences of "Webpack" in the rest of the README should be changed.

neoziro:
@s-h-a-d-o-w you are right, let's go for "webpack"!

theKashey:
Never too late 👍

theKashey:
As far `webpack` were "right" - that we are waiting for?
Links to electron forge?

Issue #884, opened 2018-03-06
Performance oriented .babelrc for TS users


theKashey:
We have to include example with babel configuration for TS users.
The key ideas:
 1. Add syntax parsers
 2. Add react-hot-loader transformation plugin
 3. Do not add any other transformation plugins to maintain build speed.

vilic:
I would suggest to use `tsconfig.json` with `module` option set to `esnext`, then have babel to transform modules with `transform-es2015-modules-commonjs`. Thus the user can take the advantages of `allowSyntheticDefaultImports` and babel interop require, so that they don't have to use import wildcard everywhere.

PetrShypila:
Does anyone know if I can target TS to `ES5` directly with `react-hot-loader`? I see in examples TS targets `ES2015` which then will be transformed to `ES5` with `babel-loader`.

theKashey:
No. In this case RHL will not _see_ any Components, ie classes, as long they will be transpile, and you will lose ability to change like a anything except ‘render’ method.

johnnyreilly:
Hey all!

Maintainer of [`ts-loader`](https://github.com/TypeStrong/ts-loader) here.  I've just been checking out v4 of this mighty fine project and discovered that with v4 I need to start using babel:

> When using TypeScript, Babel is not required, but React Hot Loader will not work without it.

With v3 I didn't need to which I liked.  Would you be able to clarify why babel is needed with v4 please?  If it's the only game in town that's cool - but if there's a way I can stick with just my TypeScript + core-js flow I'd love to.  

Either way I think I'd love to understand the limitation if you're up for explaining! 🌻

theKashey:
@johnnyreilly - so here is a fairy tale.
v4 was started to support arrow functions as class members, for the projects which does not use es5 transpilation in dev mode. Actually, result is a bit more debuggable, as long it is almost the "source" code.
So, if your component got an arrow-function member, that member will "lock" _this_, and we could not use _proxy_ as we do in v3 to do RHL's work - there is no way to _fake_ this for arrow functions.

__As result - RHL will search for classes, and inject magic method, to EVAL something in the class scope and class context__

And later we drop thousand lines from proxy, and just rely on this 3 lines long method.

If you will transpile TS to ES5 - RHL will not found any "class" it is looking for.
If you will not apply babel-plugin - RHL will be unable to update _non-prototype-based_ method. Ie "onClick" handler.

[...]

PR #996, opened 2018-05-29
Simplify TypeScript configuration

lostfictions:
As I tried to figure out how to upgrade to React Hot Loader 4 today, I was a bit confused by the options for configuring it for use with TypeScript -- it seems less than ideal to chain loaders, and in particular requiring folks to change their tsconfig is pretty annoying. Either way, there's a performance hit.

I think the ideal, most non-disruptive solution for TypeScript users would be for RHL to eventually support a TypeScript custom transform plugin as discussed [here](https://github.com/gaearon/react-hot-loader/issues/884#issuecomment-375555602) -- but for now, since Babel can consume TypeScript syntax and transpile it, why not recommend using it alone as the simplest method? There's no requirement for users to change their tsconfig, rebuilds are lightning-fast, and you can still typecheck with [fork-ts-checker-webpack-plugin](https://github.com/Realytics/fork-ts-checker-webpack-plugin) (or manually invoking `tsc --noEmit`).

In this PR I've provisionally updated the README with more verbose (but hopefully simpler) instructions and changed the TypeScript example to reflect this configuration -- maybe folks can try it out and see what they think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,82 +98,54 @@ Follow [these code examples](https://github.com/Grimones/cra-rhl/commit/4ed74af2
 
 ### TypeScript
 
-When using TypeScript, Babel is not required, but React Hot Loader will not work (properly) without it.
-Just add `babel-loader` into your Webpack configuration, with React Hot Loader plugin.
+As of version 4, React Hot Loader requires you to pass your code through [Babel](http://babeljs.io/) to transform it so that it can be hot-reloaded. This can be a pain point for TypeScript users, who usually do not need to integrate Babel as part of their build process.
 
-There are 2 different ways to do it.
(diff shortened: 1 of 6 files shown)


theKashey:
I think - its time to create a separate TS documentation file, with more deeper explanations, as long - we have got tooooo many variants.

micimize:
the preset here depends on [@babel/plugin-transform-typescript](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-typescript), which among other things doesn't support namespaces. I think the `syntax-typescript` the previous babel-only setup used has the same problem (they don't say explicitly).

For some people the performance is worth the trade-off, but it should be noted. I personally wasn't able to get a functioning typescript-first config with my environment, I think because of `allowSyntheticDefaultImports` or some other setting.

theKashey:
`allowSyntheticDefaultImports` does not change the way code got generated.
I thing we have to make a pause, and outline all possible way to handle TS, more examples to test, and all caveats found.

theKashey:
merged as a part of another PR

Issue #993, opened 2018-05-27
Allow hooking into 


natew:
### Description

I'm trying to add this to my root module:

![image](https://user-images.githubusercontent.com/12100/40590825-0a1bf79a-61bb-11e8-9d4b-ba0bb273088c.png)

Which would let me hook into a simple hydrate/dehydrate system I had setup from our previous HMR stuff. But if I add those lines, the regular HMR doesn't seem to get called.

### Expected behavior

Should have a hook for adding module.hot listeners.

### Actual behavior

Webpack seems to not allow more than one.

### Environment

React Hot Loader version: 4.2.0

Run these commands in the project folder and fill in their results:

1.  `node -v`: 9.8.0
2.  `npm -v`: 5.6.0

Then, specify:

1.  Operating system: High Sierra
2.  Browser and version: Chrome 46

Correct me if I'm wrong. I can refresh and try with and without and it seems to be the case. I think an extra hook like this would be sufficient:

```
hot(module, status => {
  // my extra hooks
})(Component)
```

theKashey:
What it the problem with this?
I mean - internally RHL is doing almost the same - https://github.com/gaearon/react-hot-loader/blob/master/src/hot.dev.js#L42
`module.hot` is not our API, and we have to power over it.

natew:
I’m just seeing that if I add my own it prevents react hot loader from
being called. So just asking for a callback function to hook into the
internal one in rhl.

Would let me patch in store state preservation logic on hmr.

theKashey:
Got the point. Let first double check webpack code, as long it probably should allow more than one status handler.

PR #1011, opened 2018-06-06
Add hook for webpack statusHandler

natew:
Probably not ideal, but would resolve: https://github.com/gaearon/react-hot-loader/issues/993#issuecomment-392380403

### DIFF
--- a/src/hot.dev.js
+++ b/src/hot.dev.js
@@ -19,7 +19,7 @@ const createHoc = (SourceComponent, TargetComponent) => {
   return TargetComponent
 }
 
-const makeHotExport = sourceModule => {
+const makeHotExport = (sourceModule, options) => {
   const updateInstances = () => {
     const module = hotModule(sourceModule.id)
     clearTimeout(module.updateTimeout)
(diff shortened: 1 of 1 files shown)


theKashey:
Do you might need before and after hooks?

Probably you need only `before`.
Probably it should be placed inside `updateInstances` to be more "not only webpack" compliant.

In case then you want to land a peace for webpack HMR I would propose another build

```js
> App.js
// addStatusHandler does not "accept" the changes, but could "observe"
module.hot.addStatusHandler(status => { do-whatever-you-want });
export default App;

> AppImporter.js
import App from './App';
// HMR will be accepted and stopped here
export default hot(App);

> index.js
import App from './AppImporter.js'
```

or, you can use anther "hack", by creating anther parent for App.js, which could observe the changes.
```js
> observer.js
import './App';  // you are _another_ parent module for the App.js.

module.hot.addStatusHandler(status => { do-whatever-you-want });
module.hot.accept(); //you have to accept the change
```

theKashey:
Also please double check will just a "file structure" solution work for you.

natew:
Adding an extra file above worked for me, thank you!

theKashey:
Glad that you are problem got solved.
Lets pause, but not close, this PR for a while - it should be a bit rethinked before the merge, as long we will not be able to revert published change.

PR #1014, opened 2018-06-09
Handle lazy indeterminate static properties

taion:


theKashey:
No tests 😭
Things like this should be covered by tests, to not to be suddenly broken in the future.

taion:
@theKashey 

I'm back with more "fun" on weird React features used in Relay.

Relay now does this: https://github.com/facebook/relay/commit/0573d773af1632dd70b03311cfc30c9b5eb4c0e6

So... we need to also lazily proxy over `getDerivedStateFromProps`. Sigh.

taion:
Sorry, tests forthcoming! Wanted to get the PR up to point people at what was happening.

taion:
Updated the test case to cover this.

theKashey:
👍

theKashey:
@taion - ok to merge? Is this the only change you need?

taion:
Yup, this is the only change. I’m not 100% sure that my use of `transferStaticProps` there is correct, but it’s probably okay. At least, it fixes the problem I see.

theKashey:
It sounds to be correct. Anyway - I should not break anything.

theKashey:
v 4.3.1

PR #1028, opened 2018-07-02
Simplify TypeScript configuration (updated)

Akuukis:
Quick work:
* took #996 as a base for working typescript example,
* upgraded from Webpack 3 to Webpack 4 and fixed configs,
* upgraded other dependencies to current newest,
* locked (remove `^`) babel dependencies because it's still in beta,
* rebased to new `master`,
* add support for `async` & `await` syntax (EDIT: amended later),
* tested and still working, faster than ever :tada:

Though I should share this. Thanks @lostfictions for heavy lifting!

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ A sample configuration:
 };
 ```
 
-For a full example configuration of TypeScript with React Hot Loader, check [here](https://github.com/gaearon/react-hot-loader/tree/master/examples/typescript).
+For a full example configuration of TypeScript with React Hot Loader and newest beta version of Babel, check [here](https://github.com/gaearon/react-hot-loader/tree/master/examples/typescript).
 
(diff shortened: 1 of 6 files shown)


theKashey:
Looking great, but many people will refuse to use beta babel 7.
Probably one should leave some message about it, as long _default_ package install will bring babel 6.

Akuukis:
Makes sense. Should we add warning to README and merge, or leave this PR until babel 7 becomes stable?

theKashey:
I have no idea when it will be done, so lets try just some notes in README

Akuukis:
@theKashey are we good to proceed?

theKashey:
All good, sorry just busy a bit in real world.

theKashey:
I've found that some TS people _do not like_ babel. They could be ok-ish to add babel as a webpack loader, but dont want to replace TS by babel as a whole.
Let me move old readme to a separate file, and point it for all non-babel-7 customers.

PR #1029, opened 2018-07-05
add code-splitting example

theKashey:
Adding example with all listed code splitting components actually being tested, and changing readme.

By fact - they all works, except `async-component` which is __completely broken__ as long stores `state` in local function-scoped variable, and we are WIPING IT! 
(so we are bad guys, killing good async-components)

@ctrlplusb  - sorry, but could you refactor https://github.com/ctrlplusb/react-async-component/blob/master/src/asyncComponent.js#L25 to be React Components state?


theKashey:
@faceyspacey - you `isHMR` function is broken! Not sure for how do you use it, but `module.hot.status` is not a thing if you are in node_modules.

_Right now_ the best way to handle HMR + RHL is to call import function in a component constructor. Not to set any states, or update some stuff - just call import, to execute a module's body and "hook into" RHL component registry.

PR #1030, opened 2018-07-06
upgrade webpack

mkg0:
This pr upgrades `webpack` version to 4 and set `react-hot-loader` version exact


theKashey:
@neoziro - what do you think about greenkeeper?

theKashey:
@mkg0 - that you for keeping us up to date!

PR #1031, opened 2018-07-07
Fix class detection

theKashey:
Actually this fixes only one problem - `isSwappable` for components with React.Component deeper in the prototype chain.
For example - StyledComponents extends another StyledComponents, which extends React.Component - our tests thinks that it's a function, as long all the things we are looking fox exists in proto-prototype.

Meanwhile I am trying to clean some duplications.

### DIFF
--- a/examples/styled-components/src/App.js
+++ b/examples/styled-components/src/App.js
@@ -20,6 +20,13 @@ const indirect = {
   ),
 }
 
+const indirectStyled = {
+  DS: styled.div`
+    border: 1px solid #f00;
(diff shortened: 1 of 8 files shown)


examples/styled-components/src/App.js
```diff
@@ -20,6 +20,13 @@ const indirect = {
   ),
 }
 
+const indirectStyled = {
```
theKashey: This was a red test

neoziro:
@theKashey looks clean!

PR #1035, opened 2018-07-26
Fix parcel example

rrousselGit:
Parcel example on the newer versions shows a 404 error on `npm start`.

Changing the start scripts to `parcel index.html` fixes the issue


theKashey:
So easy :)

PR #1038, opened 2018-07-31
Add another reason for `NODE_ENV` errors to Troubleshooting.md

Omnipotence456:
I ran into `Uncaught TypeError: Cannot read property 'NODE_ENV' of undefined` and spent a stupid amount of time debugging it. It turned out the file throwing the error had declared a function `process()` that was causing an error when the `react-hot-loader` footer tried to look at `process.env`. Sort of silly, but this might save someone time!


theKashey:
> the react-hot-loader footer tried to look at process.env

Which footer? RHL access `process` only in it's internals, and you are unable to define your own `process` there.

Issue #1041, opened 2018-08-02
Uncaught TypeError: Cannot read property 'push' of undefined


HeavenSky:
If you are reporting a bug or having an issue setting up React Hot Loader, please fill in below. For feature requests, feel free to remove this template entirely.

### Description
when I use `react-hot-loader@4` in `webpack@1` `react@0`, change a component, it will report error in browser console, and [HMR] failed. detail error is below.

What you are reporting:
```
react-hot-loader.development.js:1057 Uncaught TypeError: Cannot read property 'push' of undefined
    at hydrateLegacyStack (react-hot-loader.development.js:1057)
    at hydrateLegacyStack (react-hot-loader.development.js:1056)
    at react-hot-loader.development.js:1061
    at Array.forEach (<anonymous>)
    at hydrateLegacyStack (react-hot-loader.development.js:1059)
    at hydrateLegacyStack (react-hot-loader.development.js:1056)
    at react-hot-loader.development.js:1061
    at Array.forEach (<anonymous>)
    at hydrateLegacyStack (react-hot-loader.development.js:1059)
    at hydrateLegacyStack (react-hot-loader.development.js:1056)
```

### Expected behavior

What you think should happen: when i change component, [HMR] should success

### Actual behavior

What actually happens: [HMR] failed and reported an error in browser console.

### Environment

React Hot Loader version:  4.3.4

Run these commands in the project folder and fill in their results:

1.  `node -v`: v8.11.3
2.  `npm -v`:  6.2.0

Then, specify:

1.  Operating system:  MAC OSX 10.13.6
2.  Browser and version: chrome 67.0.3396.99

### Reproducible Demo
https://github.com/HeavenSky/webpack-frames/tree/issue/webpack1-ie8
Above is the issue demo code.
    1. start server `npm i`  `npm ddp`  `npm start`
    2. open chrome url:   http://localhost:8888/
    3. open browser console
    4. change `src/containers/App.js` component,  browser console will report error.

I try to fix this issue. edit `node_modules/react-hot-loader/dist/react-hot-loader.development.js`, LineNumber: 1057
```js
stack.children.push(childStack);  // change to `stack.children && stack.children.push(childStack);` this issue is not happened, [HMR] success.
```

Please take the time to create a new project that reproduces the issue.

You can copy your project that experiences the problem and start removing things until you’re left with the minimal reproducible demo. This helps contributors, and you might get to the root of your problem during that process.

Push to GitHub and paste the link here.

[...]

PR #1042, opened 2018-08-03
fix: some react15 trees could not be hydrated #1041

theKashey:


theKashey:
Not React15, but React14, which we do not support.

PR #1046, opened 2018-08-10
fix(hmr): check that module is not react-hot-loader

jamiehill:
# React-Hot-Loader Webpack Bug
This PR rectifies the issue as demonstrated by the follow repo:
https://github.com/jamiehill/rhl-webpack-bug

[https://github.com/gaearon/react-hot-loader/commit/ffe003532d50d7079efcac6d18254f770691f41d#diff-168726dbe96b3ce427e7fedce31bb0bc](https://github.com/gaearon/react-hot-loader/commit/ffe003532d50d7079efcac6d18254f770691f41d#diff-168726dbe96b3ce427e7fedce31bb0bc)

When using with Webpack, importing `hot` from 'react-hot-loader', the conditional is serving the production version of RHL, thus disabling HMR

DOES NOT WORK
```
'use strict'

if (!module.hot || process.env.NODE_ENV === 'production') {
  module.exports = require('./dist/react-hot-loader.production.min.js');
} else {
  module.exports = require('./dist/react-hot-loader.development.js');
}
```

### This can be demonstrated by bypassing this file altogether and importing just the development.js
See `src/App.js`.. Comment in the 'DOES WORK import to see HMR working again.

Doing a simple check for the module.filename containing 'react-hot-loader' fixes the issue.

DOES WORK
```
'use strict'

const notRHL = module.filename.indexOf('react-hot-loader') === -1
if ((!module.hot && notRHL) || process.env.NODE_ENV === 'production') {
  module.exports = require('./dist/react-hot-loader.production.min.js');
} else {
  module.exports = require('./dist/react-hot-loader.development.js');
}
```

The app now benefits from HMR, with state intact.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict'
 
-if (!module.hot || process.env.NODE_ENV === 'production') {
+const notRHL = module.filename.indexOf('react-hot-loader') === -1
+if ((!module.hot && notRHL) || process.env.NODE_ENV === 'production') {
   module.exports = require('./dist/react-hot-loader.production.min.js');
 } else {
   module.exports = require('./dist/react-hot-loader.development.js');
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "node": ">= 6"
   },
   "jest": {
+    "testURL": "http://localhost/",
     "moduleDirectories": [
       "node_modules",
       "<rootDir>"


theKashey:
Sorry, I missed this PR. Looking into now.

theKashey:
I couldn't accept your PR right now, as long `module.hot` is something RHL is rely on.
It used to determine that hot updates are supported, and it used to actually update the code inside `hot` function - https://github.com/gaearon/react-hot-loader/blob/master/src/hot.dev.js#L36

I didn't clearly remember how actually HMR works in Electron, look like it just updates everything, but your change might be a breaking change for someone, and we better search for a more accurate variables presented in electron.
(could you double check solution from https://github.com/gaearon/react-hot-loader/pull/994 ?)

theKashey:
Probably - we could inverse condition - if you are in dev - use dev, in all other cases - prod.

theKashey:
#1244 addresses the same.

PR #1054, opened 2018-08-31
fix: don't inadvertendly call getters

phyllisstein:
:wave: Hey folks! I'm submitting this patch to cover a small bug that I noticed in the RHL reconciler logic. I was working with a component that included a getter function which tried to access `this`, and hot reloading kept failing with the following error:

```
react-hot-loader.development.js:181 React-hot-loader: reconcilation failed due to error TypeError: Cannot read property 'authors' of undefined
    at PureComponent.get (post.js:69)
    at react-hot-loader.development.js:1171
    at Array.forEach (<anonymous>)
    at equalClasses (react-hot-loader.development.js:1170)
    at areSwappable (react-hot-loader.development.js:1196)
    at react-hot-loader.development.js:1430
    at Array.forEach (<anonymous>)
    at hotReplacementRender (react-hot-loader.development.js:1347)
    at next (react-hot-loader.development.js:1360)
    at react-hot-loader.development.js:1415
```

Taking a look at the code, I noticed that RHL was grabbing the prototype's property values directly, which was causing it to _execute_ the getter without a `this` context 😱. This change adjusts that logic to use `Object.getOwnPropertyDescriptor`, making the access safe and resolving any issues with unbound functions.

Thanks for considering it, and for all your hard work on RHL!

### DIFF
--- a/src/reconciler/hotReplacementRender.js
+++ b/src/reconciler/hotReplacementRender.js
@@ -81,11 +81,16 @@ const equalClasses = (a, b) => {
   let misses = 0
   let comparisons = 0
   Object.getOwnPropertyNames(prototypeA).forEach(key => {
-    if (typeof prototypeA[key] === 'function' && key !== 'constructor') {
+    const descriptorA = Object.getOwnPropertyDescriptor(prototypeA, key)
+    const valueA =
+      descriptorA && (descriptorA.value || descriptorA.get || descriptorA.set)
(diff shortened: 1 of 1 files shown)


theKashey:
Looking good and logical. We have similar logic in proxy module, for the cases like this.
Let me just double check all the things, but in my tomorrow.

PR #1058, opened 2018-09-08
change App's path in create-react-app

Laassari:


neoziro:
You have to mark your App as `hot`, so it the path is correct, you must not be in the `App.js`.

Laassari:
but `index.js` is not exporting anything, it's only responsible for `render`ing `App.js` to the DOM

theKashey:
Look like App is a right name then.

PR #1064, opened 2018-09-13
Fix babel 7 returns

theKashey:
Lets try to fix babel 7 yet agan.

What was made before - as @loganfsmyth proposed "registration" was moved from "node" in the first argument to the "state" - the second argument. Unfortunately it does not help at all.

What was made in this PR:
 - instead of Program Enter/Exit, Declaration, Class visitors I am keeping only Program Enter, and running "nested" traversal with Declaration/Class visitors only inside it.
 - in this case "registration" is a local variable in ProgramEnter visitor scope, thus it is just not possible to have the issue.

I would expect that this way is not so performant as "standard", but as long all "react-loaders" uses it to handle `import` statement before any other plugins - it's ok.

Anyway - @neoziro, @loganfsmyth - could you review the changes?

### DIFF
--- a/src/babel.dev.js
+++ b/src/babel.dev.js
@@ -82,53 +82,20 @@ module.exports = function plugin(args) {
     }
   }
 
-  const REGISTRATIONS = Symbol('registrations')
   return {
     visitor: {
-      ExportDefaultDeclaration(path, state) {
-        const { file } = state
-        // Default exports with names are going
-        // to be in scope anyway so no need to bother.
-        if (path.node.declaration.id) {
-          return
-        }
-
-        // Move export default right hand side to a variable
-        // so we can later refer to it and tag it with __source.
-        const id = path.scope.generateUidIdentifier('default')
-        const expression = t.isExpression(path.node.declaration)
-          ? path.node.declaration
-          : t.toExpression(path.node.declaration)
-        path.insertBefore(
-          t.variableDeclaration('const', [
-            t.variableDeclarator(id, expression),
-          ]),
-        )
-        path.node.declaration = id // eslint-disable-line no-param-reassign
-
-        // It won't appear in scope.bindings
-        // so we'll manually remember it exists.
-        state[REGISTRATIONS].push(
-          buildRegistration({
-            ID: id,
-            NAME: t.stringLiteral('default'),
-            FILENAME: t.stringLiteral(file.opts.filename),
-          }),
-        )
-      },
-
+      // using program to replace imports before "dynamic-import-node"
+      // see: https://jamie.build/babel-plugin-ordering.html
       Program: {
-        enter({ scope }, state) {
-          const { file } = state
-          state[REGISTRATIONS] = [] // eslint-disable-line no-param-reassign
+        enter(programPath, { file }) {
+          const registrations = []
+          const { scope, node } = programPath
 
-          // Everything in the top level scope, when reasonable,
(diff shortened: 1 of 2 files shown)


src/babel.dev.js
```diff
@@ -82,53 +82,20 @@ module.exports = function plugin(args) {
...
-        state[REGISTRATIONS].push(
-          buildRegistration({
-            ID: id,
-            NAME: t.stringLiteral('default'),
-            FILENAME: t.stringLiteral(file.opts.filename),
-          }),
-        )
-      },
-
+      // using program to replace imports before "dynamic-import-node"
```
theKashey: Left over comment :)

loganfsmyth:
Is this new behavior still throwing about `push` on `null`? Is there an example reproduction repo so I verify this behavior? I'm happy to help diagnose why the old solution didn't work. It seems like it should work.

theKashey:
Code in this PR confirmed to work well. Code in the master branch confirmed to have the same issues as before. And all examples I’ve got - working for any version.

villesau:
Do you have any vision what could cause the throwing? I could try to reduce the repro based on that. In our case: Everything worked fine in browser, and when running mocha tests. But when running our older test suite using karma/jasmine/headless chrome stack, following error happened:

```
15:30:38 [Frontend tests] fe-unit-tests_1   | Module build failed: TypeError: Cannot read property 'push' of undefined
15:30:38 [Frontend tests] fe-unit-tests_1   |     at PluginPass.ExportDefaultDeclaration (/frontend/node_modules/react-hot-loader/dist/babel.development.js:83:30)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at newFn (/frontend/node_modules/@babel/traverse/lib/visitors.js:193:21)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at NodePath._call (/frontend/node_modules/@babel/traverse/lib/path/context.js:53:20)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at NodePath.call (/frontend/node_modules/@babel/traverse/lib/path/context.js:40:17)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at NodePath.visit (/frontend/node_modules/@babel/traverse/lib/path/context.js:88:12)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at TraversalContext.visitQueue (/frontend/node_modules/@babel/traverse/lib/context.js:118:16)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at TraversalContext.visitSingle (/frontend/node_modules/@babel/traverse/lib/context.js:90:19)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at TraversalContext.visit (/frontend/node_modules/@babel/traverse/lib/context.js:146:19)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at Function.traverse.node (/frontend/node_modules/@babel/traverse/lib/index.js:94:17)
15:30:38 [Frontend tests] fe-unit-tests_1   |     at traverse (/frontend/node_modules/@babel/traverse/lib/index.js:76:12)
```

Unfortunately I'm not able to share example, but if someone could shed light to potential issue, I could perhaps try to reduce one. Otherwise we have babel 7.0.0 and webpack 3.2.0 with various plugins.

Changes in this PR works fine (assuming code is same as linked into issue).

theKashey:
We went another way.

PR #1081, opened 2018-10-23
drop `hot` calls for prod mode

### LINKED ISSUE #1080: Using `module` for `hot` cancels some webpack optimizations in prod
```js
import React from 'react'
import { hot } from 'react-hot-loader'

const App = () => <div>Hello World!</div>

export default hot(module)(App)
```

This code uses `module` variable. This cancels at least [concatenateModules](https://webpack.js.org/configuration/optimization/#optimization-concatenatemodules) optimization (enabled in prod mode)
https://github.com/webpack/webpack/blob/af90df9998c2aec0a9f8fbb6c488a0c423b1a331/lib/optimize/ModuleConcatenationPlugin.js#L104

Seems like bundles that using RHL has the worst size and eval-performance because some webpack-optimizations just not applying

I've created a simple babel-plugin, that replaces `export default hot(module)(App)` by `export default App` and it dropped **200kb** from total bundle size by dropping useless code and by ModuleConcatenationPlugin stuff.
Notу that we have only 8 places where we using `hot` in our code. 200kb for only 8 places... not bad...

But it seems like a hack... I think, that the problem should be solved on RHL side.

I can create a PR to `babel.prod.js` with replacing `hot(module)(App)` by `App`

smelukov:
closes #1080

### DIFF
--- a/src/babel.prod.js
+++ b/src/babel.prod.js
@@ -1,3 +1,64 @@
+const RHLPackage = 'react-hot-loader'
+
+function isImportFromRHL(path, name) {
+  const binding = path.scope.getBinding(name)
+  const bindingType = binding && binding.path.node.type
+
+  if (
+    bindingType === 'ImportSpecifier' ||
+    bindingType === 'ImportDefaultSpecifier' ||
+    bindingType === 'ImportNamespaceSpecifier'
+  ) {
+    const bindingParent = binding.path.parent
+    return bindingParent.source.value === RHLPackage
+  }
+
+  return false
+}
+
+function isRHLContext(file) {
+  const { modules } = file.metadata
+
+  if (modules && Array.isArray(modules.imports)) {
+    const { imports } = modules
+
+    for (let i = 0; i < imports.length; i++) {
+      const { source } = imports[i]
+
+      if (source === RHLPackage) {
+        return true
+      }
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


smelukov:
Now supports all import cases:
```
import { hot } from 'react-hot-loader'
import { hot as namedHot } from 'react-hot-loader'
import * as RHL from 'react-hot-loader'
```

theKashey:
Amazing! Just amazing!

smelukov:
Are there any comments for this? )

theKashey:
Gonna handle this and another pending PRs on weekends.

theKashey:
Released in 4.3.12

smelukov:
Thanks :)

edmorley:
Hi!

I don't see any change in the babel output when using default exports (eg `export default hot(module)(App)`). Is this expected?

smelukov:
@edmorley Hi!
The output will change only in the production build (when `NODE_ENV` is `production`)

edmorley:
Yup, this is with NODE_ENV=production, and webpack mode production.

smelukov:
@edmorley thank you! I've investigated this and find out that hot-dropping may not work with some babel-presets configuration.
Fix in https://github.com/gaearon/react-hot-loader/pull/1093

PR #1084, opened 2018-10-25
feat: React16.6 compatibility

theKashey:
## fixing
- React.memo is `invisible` to RHL
- React.memo does not support "indeterminate components" (RHL uses them for every SFC)
- React.lazy could be autoimported on variable re-registration.
- React.forwardProps (including styled-components)


src/reactHotLoader.js
```diff
@@ -16,7 +16,15 @@ import logger from './logger'
 
 import { preactAdapter } from './adapters/preact'
 
-function resolveType(type) {
+const forceSimpleSFC = { proxy: { allowSFC: false } }
```
theKashey: React.memo could "eat" only simple class or functional component.
The current approach, with "indeterminate" components is breaking the stuff.

src/reactHotLoader.js
```diff
@@ -62,9 +70,17 @@ const reactHotLoader = {
         configuration.onComponentRegister(type, uniqueLocalName, fileName)
       }
 
-      updateProxyById(id, type)
+      updateProxyById(id, type, options)
       registerComponent(type)
     }
+    if (isMemoType({ type })) {
```
theKashey: Probably the same magic is needed for Context

README.md
```diff
@@ -337,12 +337,48 @@ import { setConfig, cold } from 'react-hot-loader'
...
+
+```js
+import { setConfig, cold } from 'react-hot-loader'
+setConfig({
+  onComponentCreate: (type, name) =>
+    (String(type).indexOf('useState') > 0 ||
+      String(type).indexOf('useEffect') > 0) &&
+    cold(type),
+})
+```
```
just-boris: This code snippet and the one above are identical. Is this expected or that was supposed to be something else?

theKashey: 💩 this is something unfinished. Idea was to provide yet-another-api, to specify special flags, component should be created with.
The current approach to "fix" hooks with `setConfig({pureSFC: true})` is breaking Relay-modern.

I also could not put this logic inside, as long I could not cast every component to a string - `emotion` throws an Error and should be fixed first.

Issue #1089, opened 2018-10-28
Methods are not removed from ProxyComponent


bradennapier:
### Description

I am not sure exactly if this is expected behavior.  It causes some issues so I was not sure.  When a component is hot reloaded it will not remove any old methods that are no longer present on a Component.  

This has a few unintended side-effects.  For one, if you have a `Component` with `shouldComponentUpdate` then you change it to a `PureComponent` and remove `shouldComponentUpdate` you get an error and have to refresh the browser.

### Expected behavior

If a method is removed from a class, it should be removed form the `ProxyComponent` as well.

### Environment

React Hot Loader version: Tried on versions ranging from latest to (4.3.11) to 4.3.8.

Run these commands in the project folder and fill in their results:

1.  `node -v`: 10.12.0
2.  `npm -v`: 6.4.1

Then, specify:

1.  Operating system: OSX 10.14.0
2.  Browser and version: Chrome 70

### Reproducible Demo

Since it requires changing to see the effect this is hard but essentially starting with

#### Step One

```
export default class ExampleComponent extends React.Component<*> {
  shouldComponentUpdate() {
    console.log('Should Update?');
    return true;
  }

  test() {
    console.log('Test');
    return true;
  }

  render() {
    console.log('Rendering Example: ', this.test(), this);
    return <div>Hello</div>;
  }
}
```

![image](https://user-images.githubusercontent.com/15365418/47611705-ce07d680-da27-11e8-8f64-06a8177d2de5.png)

#### Step Two

```javascript
export default class ExampleComponent extends React.PureComponent<*> {
  render() {
    console.log('Rendering Example: ', this.test(), this);
    return <div>Hello</div>;
  }
}
```

![image](https://user-images.githubusercontent.com/15365418/47611712-f68fd080-da27-11e8-8272-5006cc225cf5.png)

#### Result

As you can see by the above, I remove the test method and I am still calling it in the `render()` and it still works great. 

Additionally, I change to using a `PureComponent` and I end up with an error/warning saying that I should not provide `shouldComponentUpdate`.  My only solution at this point is to refresh or else I get a warning every time I try anything from now on.

bradennapier:
When I think about this, I believe I remember an issue from way back when from @gaearon around why the fn's need to remain in the prototype.  Having to do with the possibility of scheduled timers and such expecting them to exist.

So perhaps we just need to detect if `shouldComponentUpdate` is removed and remove it from the ProxyComponent properly?

[...]

PR #1090, opened 2018-10-28
remove certain members if removed from prototype

bradennapier:
Adds a check on successive `proxyGeneration` iterations against a 
 `generationallyRemovedMembers` array - removing any matches if they have been removed from the component on the next generation.

This is specifically to remove `shouldComponentUpdate` and make it so the errors with changing to a `PureComponent` do not occur, however the `generationallyRemovedMembers` array is used in case other examples of this come up in the future.

See #1089


bradennapier:
Changing to a branch to fit within Contributor guidelines

PR #1091, opened 2018-10-28
Remove generational methods

bradennapier:
Adds a check on successive `proxyGeneration` iterations against a 
 `generationallyRemovedMembers` array - removing any matches if they have been removed from the component on the next generation.

This is specifically to remove `shouldComponentUpdate` and make it so the errors with changing to a `PureComponent` do not occur, however the `generationallyRemovedMembers` array is used in case other examples of this come up in the future.

See #1089 

**Note:** updated PR #1090 to do PR from branch as mentioned in Contributors guidelines.

### DIFF
--- a/src/proxy/createClassProxy.js
+++ b/src/proxy/createClassProxy.js
@@ -37,6 +37,10 @@ const blackListedClassMembers = [
   'getDefaultProps',
 ]
 
+const generationallyRemovedMembers = [
+  'shouldComponentUpdate'
+]
+
 const defaultRenderOptions = {
   componentWillRender: identity,
   componentDidUpdate: result => result,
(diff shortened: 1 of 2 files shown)


theKashey:
Sorry, this is not the right way to do it.
The right way - drop __all__ methods from a current prototype, and recreate it (https://github.com/gaearon/react-hot-loader/blob/master/src/proxy/createClassProxy.js#L216)

bradennapier:
That was my original method of doing it but I ran across an issue from @gaearon which stated that the previous methods need to be preserved incase there is something scheduled to call the previous functions.  So while removing a lifecyle method like `shouldComponentUpdate` is safe, removing all methods would not be.

I actually have a version that does remove them all as well.

bradennapier:
@theKashey I have modified it to remove all previous methods as well as added the function `filteredPrototypeMethods` that `fakeBasePrototype` also calls to get any whitelisted methods.

theKashey:
Looks good and right

bradennapier:
Yep working well on my end! Look forward to having it merged.  I’ve realized this has been causing shadow issues for my project for awhile which I never actually attributed to RHL before for whatever reason 😇

theKashey:
Unfortunately RHL has side effects :(

theKashey:
Released in 4.3.12

PR #1093, opened 2018-11-01
Support babel 7 for hot-dropping

smelukov:
Earlier, the plugin used a preset-specific stuff (e.g. `metadata.modules.import`) and this cause to hot-dropping may not work with babel 7.
Now we are working with AST directly
Moreover, default import support was dropped, cause RHL exports nothing in `default`

### DIFF
--- a/src/babel.prod.js
+++ b/src/babel.prod.js
@@ -18,7 +18,6 @@ function isImportedFromRHL(path, name) {
 
 function getRHLContext(file) {
   const context = []
-  const importSpecifiers = []
   const { body } = file.ast.program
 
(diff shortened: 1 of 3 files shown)


theKashey:
Looking much more complex, this time.

smelukov:
@theKashey it's just iterating over AST body and collects all imports declrations

smelukov:
The reason is `file.metadata.modules` was removed in babel 7 - https://github.com/babel/babel/pull/6343
This means that without this PR hot-dropping will not working with babel 7

theKashey:
Ready to merge?

smelukov:
Yep

PR #1096, opened 2018-11-02
Add sideEffects declaration to package.json

### LINKED ISSUE #1095: Consider using 'sideEffects: false' to allow webpack to tree-shake RHL
### Description

In #1081 (included in RHL 4.3.12) and #1093 (on `master`), the RHL Babel plugin now removes much of RHL when `NODE_ENV=production`, allowing webpack to concatenate modules. However parts of RHL still remain in the bundle, which could be prevented by adding `"sideEffects": false` to `package.json`:
https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

### Expected behavior

A production (`mode: 'production'` and `NODE_ENV=production`) webpack build should include as few parts of RHL as possible.

### Actual behavior

Whilst #1093 definitely helped, there is still RHL code in the bundle output:
https://github.com/edmorley/testcase-react-hot-loader-sideEffects/commit/92d39025ad45cc6350f968b10cb2b54b8d706fb8#diff-6ca863eb505df8653ce3c20b340ba580R4636

### Environment

* React Hot Loader version: `master` (83a4961b)
* `node -v`: 11.0.0
* `npm -v`: N/A (yarn 1.12.0)
* Operating system: Windows 10
* Browser and version: N/A (inspecting bundle)

### Reproducible Demo

See the repository here:
https://github.com/edmorley/testcase-react-hot-loader-sideEffects

The first commit (https://github.com/edmorley/testcase-react-hot-loader-sideEffects/commit/eb5562cce62411294cc6030b62ec29e4e58d0094) shows the output using the latest published RHL (run `yarn && yarn build > build.log` in that repo).

The second commit (https://github.com/edmorley/testcase-react-hot-loader-sideEffects/commit/92d39025ad45cc6350f968b10cb2b54b8d706fb8) was generated by `yarn link`ing in `react-hot-loader` master, then running `yarn build > build.log` again -- and shows the benefit from #1093.

The third commit (https://github.com/edmorley/testcase-react-hot-loader-sideEffects/commit/d59d4f67385f484d9c63197b342336772dd06985) shows the effect of adding `"sideEffects": false` to the `package.json` of that yarn linked `react-hot-loader` master build before running the `yarn build > build.log`.

NB: For the yarn linked builds (2nd and 3rd commit) it's very important to delete the `node_modules` in the `react-hot-loader` git clone before running `yarn build` in the testcase repo, otherwise webpack will include two copies of React (since it will treat `testcase-react-hot-loader-sideEffects/node_modules/react/` and `<RHL-git-clone>/node_modules/react/` as separate modules) - which will give the false impression the bundle size would have increased in normal usage.

edmorley:
This provides a hint to webpack and other bundlers that they may safely drop the `react-hot-loader` module if none of the imports are being used.

See:
https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

Fixes #1095.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "patch.js",
     "react-hot-loader.d.ts"
   ],
+  "sideEffects": false,
   "keywords": [
     "react",
     "javascript",


theKashey:
Sorry, but that would __not__ work. 
- react-hot-loader is actually one big side effect. API it exposes just work with side effect it already did.
- react-hot-loader uses rollup to generate a complete bundles for dev and prod, and there is __no way you can tree shake it__.
- there is no sense to "tree shake" this - https://unpkg.com/react-hot-loader@4.3.12/dist/react-hot-loader.production.min.js. Probably we should teach babel plugin to remove the rest of used RHL parts... but which parts?

edmorley:
Many thanks!

PR #1097, opened 2018-11-03
Fix react-hot-loader alias path to node_modules

r0bdiabl0:
The path in the webpack config for the react-hot-loader alias was pointing to a non-existent library which was preventing anyone who cloned this example from compiling. Changed to point to the node_module library.


theKashey:
It was pointing to local version of react-hot-loader itself. 
Just build it first.

PR #1098, opened 2018-11-07
Webpack loader

theKashey:
Reintroduce webpack loader:
- to use it on node_modules (to _cold_ everything)
- to use with typescript
- to use with react-hooks, as long they would not require _eval_ stuff we are inject via babel right now.


theKashey:
Gonna merge this, combine with React 16.6, add a bit if other fixes, and call it a day.

Issue #1088, opened 2018-10-26
[React Hooks] Incompatibility with react-hot-loader


bkniffler:
### Description

The new react hooks API does not work with react-hot-loader. Using the [example](https://reactjs.org/docs/hooks-overview.html#-state-hook) in my app with `export default hot(module)(App);` at the bottom of app.js, the following error occurs:

### Expected behavior

Should work.

### Actual behavior

```
Uncaught Error: Hooks can only be called inside the body of a function component.
```

### Reproducible Demo

Dunno of a demo is needed, just ping me if you need one.

bkniffler:
Partially related
https://github.com/facebook/react/issues/13972

simmo:
Yes I've just seen the same. I tried removing my `hot(module)(App)` equivalent  but also required removing babel plugin to get hooks to work.

theKashey:
Yes. RHL is 100% not compatible with hooks. There is just a few reasons behind it:
- SFC are being converted to Class components. There is reason - to be able to `forceUpdate` on HMR, as long there is no "update" method on SFC. I am looking for other way of forcing the update (like [this](https://github.com/facebook/create-react-app/blob/abe07e9538f0d548eefc753949a80376c9b89f8c/packages/react-dev-utils/forceUpdateHook.js#L84). So RHL is __killing SFC__.

- "hotReplacementRender". RHL is trying to do React's job, and render the _old_ and the _new_ app, to merge them. So, obviously, that's broken now.

I am going to draft a PR, to mitigate both problems. It will work, but not today.

harisvsulaiman:
This could be a workaround https://github.com/gatsbyjs/gatsby/issues/9489#issuecomment-433872202

theKashey:
Next version of RHL, with [16.6 support](https://github.com/gaearon/react-hot-loader/pull/1084) will add `memo` and `lazy` support, and thus break this "fix".
> Right now `memo`-ed components are invisible to RHL, as long they are not functions, but objects.

theKashey:
There is a more proper fix, which would work my better and ever after - [cold API](https://github.com/gaearon/react-hot-loader#disabling-a-type-change-%EF%B8%8F)

You may disable RHL for any custom type.
```js
import { cold } from 'react-hot-loader';
cold(MyComponent);
```
or just search for "useState/useEffect" inside component source code, and ❄️it.
```js
import { setConfig, cold } from 'react-hot-loader'
setConfig({
[...]

PR #1101, opened 2018-11-11
Feat: Webpack injector

theKashey:
This PR is aimed to __implement  react RFC__ https://github.com/reactjs/rfcs/pull/74

As long as there is no activity on the original RFC, and there is no way to introduce that comparison to the already published version I am going to....

Use webpack loader to inject necessary changes into the React-Dom

🙌 🙌 🙌 🙌 🙌 🙌 

And it's actually working, letting us support any React Feature from the past or the future.

### React.lazy
(build-in)It would `import` lazy automaticaly, providing auto-reloading of the changed chunks.

### React.memo
(requires patch) Now it could deep-force-update even memo-ed components, working with fiber directly. (landing additional prop on memoizedProps to bypass shallow equal)

### React.forwardRef
(requires patch) Now it could update components, working with fiber directly

### Hooks
(requires patch) Could work with hooks with `pureSFC` enabled, or `ignoreSFC` activated


neoziro:
It is **AWESOME**, I like the idea to monkey patch React 👌

Bnaya:
Dose both babel plugin and webpack loader are required for it to work?
Or just the loader?

theKashey:
- babel is required to update class-component methods
- webpack is required to support memo, and `ignoreSFC` mode, which will enable hooks.

theKashey:
Lets discuss all the problems in https://github.com/gaearon/react-hot-loader/issues/1088

PR #1105, opened 2018-11-20
Update Config type declaration with pureSFC and allowSFC properties

BabkinAleksandr:


theKashey:
Sorry, that's my bad, but this change is a bit obsolete - https://github.com/gaearon/react-hot-loader/pull/1101/files#diff-f464595d40b58d1b6989f0cd0b5f2229

BabkinAleksandr:
That's OK. I just had typescript error while using react-hot-loader with react hooks. So, problem is solved, that's what I need.

theKashey:
Have a good day then :)

Issue #1109, opened 2018-11-25
Babel plugin causes react to be added to all Webpack outputs even if not used


driskell:
If you are reporting a bug or having an issue setting up React Hot Loader, please fill in below. For feature requests, feel free to remove this template entirely.

### Description

Granted this can be edge case as most of the time all outputs from a webpack will contain react. In a couple of my projects, however, we use a single webpack and build many bundles for many different areas of the application, both frontend and backend, and also to generate render blocking bundles separate to deferred bundles - and react is always in a deferred bundle.

With multiple entry points in webpack, some using React and some not, as described above, all outputs contain react, react.production.min and the react hot loader code, in some cases tripling or quadrupling the size of what should be small outputs.

### Expected behavior

Outputs that do not require react, do not have react hot loader in them or any of its dependencies.

### Actual behavior

The bottom of the output always contains the following, even if the output never had react within it, and therefore pulls in additional chunks of code. This is added by the babel plugin.

```
(function () {
    var reactHotLoader = __webpack_require__(0).default;

    var leaveModule = __webpack_require__(0).leaveModule;

    if (!reactHotLoader) {
        return;
    }

    reactHotLoader.register(_extends, "_extends", "non-react-import.js");
    reactHotLoader.register(_typeof, "_typeof", "non-react-import.js");
    leaveModule(module);
})();
```

The following are therefore added to the bundle unnecessarily (as can be seen in the source map):

```
"webpack:///./node_modules/react-hot-loader/index.js",
"webpack:///(webpack)/buildin/harmony-module.js",
"webpack:///./node_modules/react/index.js",
"webpack:///./node_modules/fbjs/lib/emptyFunction.js",
"webpack:///./node_modules/react-hot-loader/dist/react-hot-loader.production.min.js",
"webpack:///./node_modules/object-assign/index.js",
"webpack:///./node_modules/fbjs/lib/emptyObject.js",
"webpack:///(webpack)/buildin/module.js",
"webpack:///./node_modules/react/cjs/react.production.min.js",
```

As an example, it increases a 156K bundle to 170K, an increase of 14K. For small bundles that are 10K it's a doubling in size. Granted, this is not a significant, but it was unexpected and took time to investigate what the impact was as we hadn't anticipated this code to be in the bundle at all.

### Environment

React Hot Loader version: 4.3.12

Run these commands in the project folder and fill in their results:

1.  `node -v`: 8.12.0
2.  `npm -v`: 6.4.1

Then, specify:

[...]

PR #1119, opened 2018-12-10
Make react-hot-loader optimise away when DefinePlugin sets NODE_ENV

driskell:
Related to #1109 and likely other ones.

Existing behaviour is that babel compilation is different depending on the NODE_ENV of the webpack process, removing hot() calls when production. However, there is very little documentation to set this that I can find. It also seems strange that this seems to be the only thing that refers to NODE_ENV in webpack - everything else is impacted by the DefinePlugin instead in generated code.

This PR replaced the two types of generated code with a single generated code that can be optimised by DefinePlugin/Uglify into the correct output desired for production usage.

With Uglify 3 and the imports tree shaking in Webpack 4 it should produce fully clean output. With Uglify 2 and Webpack 3 it will product the same output as currently - which is that react-hot-loader is not fully dropped from production due to the tree shaking not working on unused imports (sideEffects is not supported).

Wondering what people's thoughts are on this approach.


src/babel.js
```diff
@@ -27,39 +29,63 @@ module.exports = function plugin(args) {
...
     templateOptions,
   )
   const headerTemplate = template(
-    `(function () {
-       var enterModule = require('react-hot-loader').enterModule;
-       enterModule && enterModule(module);
-     }())`,
+    `
+(function () {
+  // Allow tree shaking to drop the react-hot-loader module by wrapping with a conditional that DefinePlugin can replace
```
theKashey: lets move this comment above template, or you will add it to the every file processed.

src/babel.js
```diff
@@ -27,39 +29,63 @@ module.exports = function plugin(args) {
...
+  // This cannot be a ternary - must be an IF
+  if (${processEnvNodeEnv} != 'production') {
```
theKashey: `!==`, `===` - should be a habbit

theKashey:
You have to regenerate jest snapshots

theKashey:
I am not sure yet. Lets first update jest snapshots, so we would see the result code.

Issue #1111, opened 2018-11-30
Example Error Boundary


BetterCallSky:
### Description

> React supports error handling out of the box since v16 using componentDidCatch. You can create your own Error Boundary and install it after hot has been applied:

Is it possible to provide an example Error Boundary?

I tried to create my own component using the official example with 'redbox-react', but the error is not gone after doing more changes. I keep the error in the state, and the error is never cleared.

Example component
```tsx
import React from 'react';
import RedBox from 'redbox-react';

export class ErrorBoundary extends React.Component<any, any> {
  constructor(props: any) {
    super(props);
    this.state = { hasError: false };
  }

  static getDerivedStateFromError(error: Error) {
    // Update state so the next render will show the fallback UI.
    return { hasError: error };
  }

  componentDidCatch(error: Error, info: any) {
    console.error(error, info);
  }

  render() {
    if (this.state.hasError) {
      return <RedBox error={this.state.hasError} />;
    }

    return this.props.children;
  }
}

```

How to clear the state if there are any changes?

### Expected behavior

Flow should look like this:
 edit code -> runtime error -> error boundary (red box) -> edit code again and fix error -> error boundary is hidden.

### Actual behavior

edit code -> runtime error -> error boundary (red box) -> edit code again and fix error -> error boundary is still visible.

### Environment

React Hot Loader version: 4.5.1

theKashey:
:nods: It's not easy to catch "update" event and remove error overlay. 
@neoziro - I think we have to make ErrorOverlay a part of AppContaner

theKashey:
Never mind. AppContainer has a prop named `errorReporter`, you just __can`t__ provide a settings for it :)

BetterCallSky:
@theKashey 

Should I also use`<AppContaner />`? In readme, it's mentioned to use only `hot(..)`.  Here https://github.com/gaearon/react-hot-loader#migrating-from-v3

theKashey:
In short:
 - the feature you are asking for is supported
 - you can't use it :)

We have to:
 - create a way to specify settings to the AppContainer created __inside__ hot
 - probably add one more field to `setConfig`, as long as your settings should be applied application wide.

theKashey:
4.6.0 gives you everything.
Check `errorReporter` and `ErrorOverlay` options of `setConfig`

BetterCallSky:
@theKashey thanks! 

[...]

Issue #1078, opened 2018-10-08
Hot reloading shouldn't stop after an error


mjhoffm2:
### Description

Currently, if there is a compilation error in the code being hot reloaded, I will see an overlay containing information about the errors, as well as a message in the console about these errors.  However, after these errors are addressed, hot reloading should continue as normal, but currently isn't.

### Expected behavior

When I save a piece of code with an error, then fix the error, I should be able to continue editing and hot reloading my application.  This works perfectly fine without using react-hot-loader, and just using webpack-hot-middleware.

```ts
function render(rootContainer: JSX.Element) {
    ReactDOM.render(
        <Provider store={store}>
            <ConnectedRouter history={history}>
                {rootContainer}
            </ConnectedRouter>
        </Provider>,
        document.getElementById('root')
    );
}

render(<Routes />);

//configure hot module replacement during development
if(module.hot) {
    module.hot.accept('./routes', () => {
        const NewRoutes = require('./routes').Routes;
        render(<NewRoutes />);
    });
}
```

![image](https://user-images.githubusercontent.com/29382626/46637642-bb823780-cb22-11e8-82d2-22229a571722.png)

### Actual behavior

What actually happens:

When using react-hot-loader with `hot(module)( ... )` in my Routes component, my application is no longer able to hot reload after I fix the error.  I see the following:
![image](https://user-images.githubusercontent.com/29382626/46637458-e029df80-cb21-11e8-9198-ad05457d4f87.png)

I see the error overlay disappear, but my component does not get updated any more.

### Environment

React Hot Loader version: 4.3.11

Run these commands in the project folder and fill in their results:

1.  `node -v`: v8.10.0
2.  `npm -v`: 6.4.1

Then, specify:

1.  Operating system: Windows 10
2.  Browser and version: Google Chrome Version 69.0.3497.100 (Official Build) (64-bit)

### Reproducible Demo

I can do if necessary

theKashey:
Yeah - it would be great to have a demo.
According to your logs - RHL just didn't setup a new `module.hot.accept`, and that's strange.

mjhoffm2:
Here is a demo: https://github.com/mjhoffm2/react-demo/tree/2bbdb5bc95e4f322c41ea6ca3150b43309286a81/node

[...]

PR #1122, opened 2018-12-11
New `hot` API

theKashey:
This PR is aimed to solve 2 issues:
- there no way to configure `errorReporter` using hot - there were several issues about it, like #1111 
- in case of JS error out of RHL scope we are not charging a new `module.hot.accept` and that's lead to full page reload next time - #1078

## The change
- split `hot` into two pieces
```js
import {hot} from 'react-hot-loader/root';
//^^ this will setup HMR for the "parent"(current) module, and then wipe itself from a cache, to execute the same action for the next call

export hot(Component);
// ^ this call is a bit shorter, and does not use magic `module` variable.
```

If "first" part is called, but second is not - then some JS error took a place, and we should display an Error overlay.


neoziro:
I think it is very nice to simplify API and to remove the magic "module" part! Good job!
I can't review it in detail because I no longer understand the whole code.

PR #1128, opened 2018-12-15
Initialize error overlay when DOM is loaded

### LINKED ISSUE #1127: document.body may not be created yet when initializing RHL
If you are reporting a bug or having an issue setting up React Hot Loader, please fill in below. For feature requests, feel free to remove this template entirely.

### Description

With RHL 4.6, when I can not load my entrypoint with a script tag in my page's `<head>`.

### Expected behavior

Everything should work

### Actual behavior

JS error: `Uncaught TypeError: Cannot read property 'appendChild' of null` on this line: https://github.com/gaearon/react-hot-loader/blob/f15b108cfac2aa6b8f98496b1e3103a4ad1c9e9c/src/errorReporter.js#L112

### Environment

React Hot Loader version: 4.6.0
Chrome 71

### Analysis

The `<script>` tag is synchronous, so it blocks the page loading to download the JS file and executes it. But as this is done while processing the document's `<head>`, the `<body>` tag is not yet processed, thus there is no `<body>` node in the page DOM yet.

### Possible fix

Fix wrap this function in `document.addEventListener("DOMContentLoaded", …)` so it is executed once the whole document has been parsed and the `<body>` tag exists.

### Workaround

Until this is fixed, you can move your `<script>` tag inside `<body>` (at the start or at the end) and everything will work fine

renchap:
Initialize error overlay when DOM is loaded, otherwise `document.body` may not exist yet.

My only change is wrapping `initErrorOverlay` in `document.addEventListener('DOMContentLoaded')`.

Fixes #1127 (more details here).

I tested this with my app (using `yarn link`) and it fixed the error when the `<script>` tag is in `<head>`.

### DIFF
--- a/src/errorReporter.js
+++ b/src/errorReporter.js
@@ -105,17 +105,27 @@ class ErrorOverlay extends React.Component {
 }
 
 const initErrorOverlay = () => {
-  let div = document.querySelector('.react-hot-loader-error-overlay')
-  if (!div) {
-    div = document.createElement('div')
-    div.className = 'react-hot-loader-error-overlay'
-    document.body.appendChild(div)
+  const doInit = () => {
+    let div = document.querySelector('.react-hot-loader-error-overlay')
+    if (!div) {
+      div = document.createElement('div')
+      div.className = 'react-hot-loader-error-overlay'
+      document.body.appendChild(div)
+    }
+    if (lastError.length) {
+      const Overlay = configuration.ErrorOverlay || ErrorOverlay
+      ReactDom.render(<Overlay errors={lastError} />, div)
+    } else {
+      div.parentNode.removeChild(div)
+    }
   }
-  if (lastError.length) {
-    const Overlay = configuration.ErrorOverlay || ErrorOverlay
-    ReactDom.render(<Overlay errors={lastError} />, div)
+
+  // document.body may not exist yet if the DOM is not fully loaded
+  // this can happen if this is loaded from a sync script in <head>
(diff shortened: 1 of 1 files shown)


renchap:
After some experiments, this might break things as `initErrorOverlay` may be called multiple times and `DOMContentLoaded` is only called once.
I changed my PR to only defer the init after `DOMContentLoaded` if `document.body` does not exist, otherwise it runs immediately.

relign:
If the code is running inside a web worker, document is undefined, how to do ?? @renchap

src/errorReporter.js
```diff
@@ -105,17 +105,27 @@ class ErrorOverlay extends React.Component {
...
+    }
   }
-  if (lastError.length) {
-    const Overlay = configuration.ErrorOverlay || ErrorOverlay
-    ReactDom.render(<Overlay errors={lastError} />, div)
+
+  // document.body may not exist yet if the DOM is not fully loaded
+  // this can happen if this is loaded from a sync script in <head>
+  if (document.body) {
+    doInit()
```
relign: web worker don't have document

theKashey:
Sorry for a delay - have experienced a power outage after the storm.
Anyway - this problem could be solved in a more easy way, and I already have in a local branch.
just `if(!document)return`.

theKashey:
Already testing... Unfortunately there is no e2e tests here, so testing is partially manual :(

renchap:
Closing as this has been fixed by another PR

PR #1155, opened 2019-01-17
Use function component instead of class in prod

### LINKED ISSUE #1154: Use function component instead of class in prod build?
I noticed that the prod build is currently 1.43 KiB. https://unpkg.com/react-hot-loader@4.6.3/dist/

It looks like most of the weight comes in because the AppContainer component is a class

https://github.com/gaearon/react-hot-loader/blob/015fcf09ff35e0bfec7e8d0702ce4419f5966d0a/src/AppContainer.prod.js#L5-L9

Would it make sense to change this to be a function component instead to reduce production bundle size?

```jsx
function AppContainer(props) {
  return React.Children.only(props.children)
}
```

lencioni:
I noticed that the prod build is currently 1.43 KiB.
https://unpkg.com/react-hot-loader@4.6.3/dist/

It looks like most of the weight comes in because the AppContainer
component is a class

https://github.com/gaearon/react-hot-loader/blob/015fcf09f/src/AppContainer.prod.js#L5-L9

We can avoid most of this weight by using a function component instead
of a class here.

Fixes #1154

### DIFF
--- a/src/AppContainer.prod.js
+++ b/src/AppContainer.prod.js
@@ -2,10 +2,8 @@
 
 import React from 'react'
 
-class AppContainer extends React.Component {
-  render() {
-    return React.Children.only(this.props.children)
-  }
+function AppContainer(props) {
+  return React.Children.only(props.children)
 }
 
 export default AppContainer


theKashey:
👍

PR #1186, opened 2019-02-20
Avoid using react deprecated Context.Consumer.Provider api

eyezak:
Recent changes in React 16.8.2 deprecates accessing Context.Consumer.Provider, printing warnings for each use case that does so.

This patch attempts to detect a Consumer using types before accessing the provider through the _context property. There may be better approaches here

### DIFF
--- a/src/reactHotLoader.js
+++ b/src/reactHotLoader.js
@@ -6,6 +6,7 @@ import {
   isMemoType,
   isForwardType,
   isContextType,
+  isContextConsumer,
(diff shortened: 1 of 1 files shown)


theKashey:
Sounds legit. Let me test some edge cases first.

theKashey:
Closing in favor of more robust PR https://github.com/gaearon/react-hot-loader/pull/1190
Your one was or for `const {Context, Provider} = createContext()` but broke simple `const context = createContext`

PR #1187, opened 2019-02-21
Update install instructions in README.md

kengreeff:
Hey Guys,

I had a bit of trouble understanding which `hot` was the new one when first reading the docs so I rearranged it (not sure if it is the right way round though!)

I assumed that `import { hot } from 'react-hot-loader/root'` is the correct version as it was documented as step 2.

Also made small changes to make it a bit easier to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,19 +66,19 @@ const App = () => <div>Hello World!</div>
 export default hot(App)
 ```
 
-There is also old version of `hot`, used prior version 4.5.4. **Please use a new one**,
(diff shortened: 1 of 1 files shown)


theKashey:
screen shot 2019-02-21 at 1 10 24 pm
Could you place something like `### Old API` before the note, to separate it from a instalation steps.

theKashey:
Thanks mate! Don't hesitate to improve other sections.

PR #1193, opened 2019-02-27
chore(package): Bump hoist-non-react-statics major

eps1lon:
Bump was major because of dropped `react@0.13` support which was reintroduced in `hoist-non-react-statics@3.2`. `hoist-non-react-statics@2.x` does not support forwardRef.

[hoist-non-react-statics CHANGELOG](https://github.com/mridgway/hoist-non-react-statics/blob/master/CHANGELOG.md)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3332,10 +3332,12 @@ hoist-non-react-statics@^2.3.1:
   resolved "https://registry.yarnpkg.com/hoist-non-react-statics/-/hoist-non-react-statics-2.3.1.tgz#343db84c6018c650778898240135a1420ee22ce0"
   integrity sha1-ND24TGAYxlB3iJgkATWhQg7iLOA=
 
-hoist-non-react-statics@^2.5.0:
-  version "2.5.0"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -5571,21 +5573,26 @@ rc@^1.1.7:
     minimist "^1.2.0"
     strip-json-comments "~2.0.1"
 
-react-dom@^16.6.3:
-  version "16.6.3"
-  resolved "https://registry.yarnpkg.com/react-dom/-/react-dom-16.6.3.tgz#8fa7ba6883c85211b8da2d0efeffc9d3825cccc0"
-  integrity sha512-8ugJWRCWLGXy+7PmNh8WJz3g1TaTUt1XyoIcFN+x0Zbkoz+KKdUyx1AQLYJdbFXjuF41Nmjn5+j//rxvhFjgSQ==
+react-dom@16:
```
eps1lon: Looks like some noise from a different yarn version. I used `yarn@1.13.0`. It's using the exact string from the package.json not just a matching semver range.

theKashey:
Keep your dependencies up-to-date :)

PR #1214, opened 2019-03-27
Pass defined falsy values in consumers' children in replacement render

aberezkin:
I found out that there was a problem with context that provide a **defined falsy** values such as `false`. If the provided value is falsy in general `hotReplacementRender` will just provide consumers' children with plain `undefined`.

It can lead to various problems. In my case it was `PropTypes` logging required errors in the console, but it can lead to more serious inconveniences if you write a component that expects context variable to be defined.

In this PR I made `hotReplacementRender` check if provided value is `undefined` and replace it with `childType[CONTEXT_CURRENT_VALUE]` only if this is the case.

I have a gut feeling this may be a consequence of a more complex bug, because it seems that `childType[CONTEXT_CURRENT_VALUE]` should contain the last context value before hot update but in my case it did not.

Thanks for your attention in advance.

### DIFF
--- a/src/reconciler/hotReplacementRender.js
+++ b/src/reconciler/hotReplacementRender.js
@@ -287,10 +287,12 @@ const hotReplacementRender = (instance, stack) => {
         next(stackChild.instance)
       } else if (isContextConsumer(child)) {
         try {
+          const contextValue = stackContext().get(getContextProvider(childType))
           next({
             children: (child.props ? child.props.children : child.children[0])(
-              stackContext().get(getContextProvider(childType)) ||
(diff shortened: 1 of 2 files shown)


theKashey:
Looking good. It’s probably worth to add a test for this case/check default react behaviour.

aberezkin:
@theKashey, looks like there was no tests for context at all so I added some tests just to prevent degradation in this area. And the one with description `passes provider defined falsy context value on receiving new children` fails without the changes introduced in this PR.

theKashey:
👍 💯 🥇

PR #1217, opened 2019-03-30
Fix Uncaught SyntaxError: Unexpected token ) on Chrome 55

izorg:


theKashey:
😨

theKashey:
Prettier is adding it, expecting than build step will remove it back

PR #1218, opened 2019-04-01
Introduce react-hot-reload.macro

cometkim:
Hi forks.

This PR simply introduced [react-hot-reload.macro](https://github.com/cometkim/react-hot-reload.macro) as an alternative option in README, without ejecting section.

It has limitations(only support `hot()`) but I believe it's the simplest way to use RHL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,8 @@ export default hot(module)(App)
 
 Users [report](https://github.com/gaearon/react-hot-loader/pull/729#issuecomment-354097936), that it is possible to use [react-app-rewire-hot-loader](https://github.com/cdharris/react-app-rewire-hot-loader) to setup React-hot-loader without ejecting.
 
+Also, [react-hot-reload.macro](https://github.com/cometkim/react-hot-reload.macro) provides limited support of React Hot Loader in zero configuration.
+
 ### TypeScript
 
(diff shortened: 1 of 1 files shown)


theKashey:
😢 it does nothing, only removing `hot` in production mode.
Unfortunately using RHL without babel or webpack plugin could lead to extreme _WTF_ situations due to internal mechanics, which would  _merge_ updated components using duck typing.

For example - all lazy-loaded components (using code splitting library, not React.Lazy) - are the same - same display name, and the same code - difference is in state/props. So - RHL would merge them.

The only way to prevent it - _register_ top level variables, but who will do it :(

cometkim:
Thanks @theKashey , I didn't realize my macro doesn't traverse for RHL, to exec`plugin()` only returns its visitor.

And you totally right, babel macro seems to be an invalid approach at this moment because it doesn't have Program scope for registration...

cometkim:
I'm not sure if this is a valid approach (maybe depends on the goal of babel macro), but it will be very convenient if it is. so I will continue to look for it.

Any feedback will be appreciated.

theKashey:
I am encouraging you to try _less_ babel ways, when it just not required, or _more_ babel ways, when it could help more than today.

Honestly - react-hot-loader probably is doing something wrong. Please, don't just accept our decisions, they might be wrong.

PR #1221, opened 2019-04-05
fix: Allow refs on lazy components

eps1lon:
See facebook/react#15318 for full context

### DIFF
--- a/src/reconciler/fiberUpdater.js
+++ b/src/reconciler/fiberUpdater.js
@@ -19,11 +19,11 @@ export const updateLazy = (target, type) => {
         // chunks has been updated - new hot loader process is taking a place
         enterHotUpdate()
         return {
-          default: props => (
+          default: React.forwardRef((props, ref) => (
             <AppContainer>
-              <C {...props} />
+              <C {...props} ref={ref} />
             </AppContainer>
-          ),
+          )),
         }
       })
     target[lazyConstructor].isPatchedByReactHotLoader = true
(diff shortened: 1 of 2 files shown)


test/AppContainer.dev.test.js
```diff
@@ -927,6 +928,7 @@ describe(`AppContainer (dev)`, () => {
       await Promise.resolve(1)
 
       expect(spy).not.toHaveBeenCalled()
+      expect(ref).toHaveBeenCalledTimes(1)
```
eps1lon: When running only this test it worked but when running all it fails. Those things should probably be wrapped in `act` to work properly. I don't have enough context here to make a good judgement how the full test should work with regards to timing.

src/reconciler/fiberUpdater.js
```diff
@@ -18,12 +18,21 @@ export const updateLazy = (target, type) => {
         const C = resolveType(m.default)
         // chunks has been updated - new hot loader process is taking a place
         enterHotUpdate()
+        if (!React.forwardRef) {
```
eps1lon: I thought this path was only used for `React.lazy` which is a 16.6 feature where React.forwardRef should be available. But react 15 uses this path as well which caused https://travis-ci.org/gaearon/react-hot-loader/jobs/516138894#L790

theKashey:
Excellent! Thank you good context and a quick fix.

theKashey:
Released

Issue #1241, opened 2019-05-07
Use `semi`


theKashey:
Long story short - by `react-hot-loader` is the only project I am more or less responsible for which uses `no semicolons` rule.

I am pretty sure that about a year ago the TC39 make it clear - automatic semicolon insertion is an evil, and encourages us to use them - https://github.com/tc39/ecma262/pull/1062

@neoziro - look like you are not using `semi` in `smooth-code` projects, so I need your opinion in this.

`semi:false` here is some sort of legacy. __Should RHL start `semi:true`__?

neoziro:
It is OK for me!

theKashey:
:sweet:, I’ll reconfigure prettier then.

PR #1242, opened 2019-05-09
migrate to semicolon-ish code style

theKashey:
Implements #1241 

- semicolons are recommend by TC39 and _future proof
- automatic semicolon injection is a dangerous thing and we had to use it for root (untranspiled) files to prevent runtime exception (shipped once a broken version to prod).
- original package creator was more about `semi:true` and `react` itself uses `semi`
- this is the only package under my governance, which does not use `semi`, and it was always a bit painful for me to fight with my coding habbits (and another maintainer approved the changes)


JustFly1984:
just why?

theKashey:
@JustFly1984 - updated description.

JustFly1984:
I have a question: can I use Typescript without semicolons? The other issue I have with semicolons is that it will point git blame on the author who inserted a semicolon, instead of original author of the code. same issue with last comma in the list.

theKashey:
> can I use Typescript without semicolons?

It would be much more safe to use TS without semi than JS. Still there are valid, IIFE-like cases when you have to insert semicolon thoughtfully, and no linter/language could help you.

> The other issue I have with semicolons is that it will point git blame on the author who inserted a semicolon

Usually it's not a problem. You are inserting a line as a complete thing - with(or without) semicolon since the beginning. 

If you are pointing on changing _blame_ for RHL source codes - this is not a first project-wide transformation we did, and (I hope) _blame_ attribution is not so important for all the people contributed into RHL.

theKashey:
Anyway, let's confirm the last sentence.

theKashey:
So 🤷‍♂️- nobody object

PR #1244, opened 2019-05-09
Fix development version in production bundle (Parcel)

doasync:
> When I use parcel build to generate a production bundle for my project that uses react-hot-loader, the bundle ends up containing react-hot-loader.development.js. However I don't want the actual hot reloading code in my production bundle, and react-hot-loader is supposed to omit this when built for production.

https://github.com/parcel-bundler/parcel/issues/1593

Screenshot from 2019-05-10 00-35-25

This PR fixes the issue just by changing the order of production checks

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
 
 var hasWindow = typeof window !== 'undefined';
 
-if (!module.hot || process.env.NODE_ENV === 'production' || !hasWindow) {
+if (process.env.NODE_ENV === 'production' || !module.hot || !hasWindow) {
   module.exports = require('./dist/react-hot-loader.production.min.js');
 } else {
   var evalAllowed = false;


theKashey:
> This PR fixes the issue just by changing the order of production checks

:tableflip:, are you sure it works as expected? Why I might not worked before?

theKashey:
Let's make it more like https://github.com/parcel-bundler/parcel/issues/1593#issuecomment-491092870
Should be more future proof and may provide better feedback for a user.

theKashey:
Implementing this in a form I've proposed in a linked parcel issue. 
Going to close this PR and thank you for shedding light on this.

doasync:
@theKashey, thanks

theKashey:
shipped in v4.8.5

PR #1250, opened 2019-05-13
fix(error): Can't find variable: require [react-static]

dalisoft:
This [bug](https://github.com/nozzle/react-static/issues/1164) is caused by something, but referring to this line and i'm fixed (should work now)

### DIFF
--- a/src/webpack/webpackTagCommonJSExports.js
+++ b/src/webpack/webpackTagCommonJSExports.js
@@ -3,6 +3,9 @@
 
 (function register() {
   // eslint-disable-line no-extra-semi
+  if (typeof require === 'undefined') {
+     return;
+  }
   /* react-hot-loader/webpack */
   const reactHotLoader = require('react-hot-loader').default;
 


dalisoft:
I can confirm the bug fixed after modifying this line

theKashey:
Could you double check - is there any file where `require` is not _undefined_, or for which files it is.

Probably this should be something like
```js
const r = typeof require !== 'undefined' ? require : (typeof __webpack_require__ !== 'undefined' ? __webpack_require__ : null);
if(!r) return;
```

But first we have to understand _why_ you have this problem.

dalisoft:
The problem is causing by react-hot-loader within react-static. I don’t have fully controllable source code

theKashey:
But we have to understand why you are the first one, who run into this issue.
@tannerlinsley - I heard you had some good time with React-Hot-Loader, so have you seen something like this before?

J-F-Liu:
Does this mean that react-hot-loader will have no effect in react-static?

theKashey:
No, disabling `webpack-loader` would lead to too hot reloading, with many false positives. Something should bound "things" to modules and their names.

tannerlinsley:
I am pretty sure this happens when trying to dynamically `import()` a module during hot-module enabled development. Can anyone confirm?

theKashey:
#1247 would solve this problem using `reactHotLoaderGlobal` variable.

PR #1253, opened 2019-05-17
fix: No HMR missing error in SSR

xD3CODER:
The second condition is triggered in a server environment resulting in the error message 
`React-Hot-Loader: Hot Module Replacement is not enabled`
Invert the two condition fixes this behavior


theKashey:
😸much simpler than I thought :) I was going to change `hot`, not index.

Issue #1138, opened 2018-12-21
Reversed Hot-Replacement


theKashey:
parent issue: #1121 
is an action for: #1024 

There are huge problems updating Class-based components. We fight with them for a year, and look like there is only one way to defeat them - surrender.

### How it works currently
- RHL set prototype from a `new` class to an `old` class
- As long as the class constructor could not be "updated" or "redone" we have to  "find" updates manually
- RHL creates `new` and `old` classes, looks for differences (not all of them are visible)
- __overrides__ class variables to replicate changes on the real instances (not all changes are replicatable) using `eval` injected by the babel plugin to maintain the "variable scope" (which might be wrong)

In short - it __lands__ some important "changes" from hot-updated code to the existing ones. Maintaining the state just because it is not updated. The result could be far from reality.

- the prototype-based methods are all "new", they are all right
- any other method update is )transported_ from a new (variable) scope to an old one
- some operations, performed in a constructor, could not be redone and skipped

### Proposal
- create a new component from a scratch, as we are doing
- detect any difference between the `old` component and instance, as we are going
- inject all changes we want to __keep__ to the new instance
- replace old instance by a new one on fiber level.
- update refs if they existed (probably could be done on createElement level)

In short - __keeps__ some important changes, like state or lifetime variables. The result is a reality, the same as if you refresh a page.

This is actually a _refresh_, but with a state kept (and we don't need anything else)

The only difference - we should inject a new instance deeply to `react-dom`.

- 👍  react 16 only
- 👍 babel plugin is __not__ required, as long there is no need to inject updated code to existing classes.
- 👍 any third party code, like [`react-redux` should be hot-reloadable](https://github.com/reduxjs/react-redux/pull/1137) without any change from their side.
- 👍 no more type comparison problem

- 👎 react-🔥-dom is required

It might be a blocker. The more we patch sources of react, the deeper we are in trouble. This this proposal is something, that _should be done_, I am not sure that it's something, that _could be done_.

theKashey:
Superseded by #1261

theKashey:
#1261 does not solve class reloading problem, this idea is still valid and required.

PR #1261, opened 2019-05-31
feat: transparent class wrapping

theKashey:
WIP
- moves class wrapper to a fiber level removing a parasite side effect from a user level.

This is a major milestone for __v5__ and removes one of the biggest RHL's sideeffect type comparison - #304

### State before
`hot-loader/react-dom` introduces controller type comparison. We injected a few things to `react-dom`
- `hotCleanup`, to cleanup hooks rendering
- `hotRenderWithHooks`, to be able to render hooks in `hotReplacementRender`
- `setHotElementComparator`, to control type comparison, injected into `reconcileSingleElement`

### New
This PR adds one more injection
- `setHotTypeResolver` injected into `createFiberFromTypeAndProps`

### How it works
- no more user-space proxying
- Elements are passed to as-is, so `<Type/>.type==Type` both for `FC`(worked before), and now for `Classes`
- On fiber creation we are replacing the original `type` but a `proxy`
- `hotElementComparator`, which compares `fiber` node with `element` now compares `proxy` with the original type, and have to `wrap` element to perform comparison.

This is a replacement for #1138 

### Left to do
- changing hook order still throws an error
- classes are still could not be fully updated due to unrepeatable constructors.

I recon the second issue could be if not skipped, then deprecated due to the hooks.

### DIFF
--- a/src/reactHotLoader.js
+++ b/src/reactHotLoader.js
@@ -117,53 +117,61 @@ const reactHotLoader = {
       configuration.showReactDomPatchNotification = false;
       // console.warn('react-🔥-loader activated.');
     }
-    /* eslint-enable */
-    if (!React.createElement.isPatchedByReactHotLoader) {
-      const originalCreateElement = React.createElement;
-      // Trick React into rendering a proxy so that
-      // its state is preserved when the class changes.
-      // This will update the proxy if it's for a known type.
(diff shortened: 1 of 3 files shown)


theKashey:
@gaearon - I need your eye on this. This mitigates the major RHL issue, but injects more and more code into react internals.

PR #1288, opened 2019-07-06
Replace lodash/merge with a very simple deep merge

### LINKED ISSUE #1269: Using lodash-es
If you configure a lodash alias in webpack:

```
config.resolve.alias['lodash'] = path.dirname(require.resolve('lodash-es'))
```

RHL fails with:

```
Uncaught TypeError: merge is not a function
    at updateProxyById (react-hot-loader.development.js:1089)
    at Object.register (react-hot-loader.development.js:2533)
```

Because `const merge = require('lodash-es/merge')` results in `{ default: merge }`.
Haven't figured out how to workaround this yet.

KidkArolis:
This is by no means an equivalent function, but I'm just wondering - isn't this enough?

In fact, maybe the `updateProxyById` code could be restructured to avoid the need to deep merge in the first place? I'm not familiar with the code enough to know if this merge replacement is sufficient to merge `renderOptions` and `componentOptions`.

Note, even arrays are merged differently using this code:

```js
var object = {
  'a': [{ 'b': 2 }, { 'd': 4 }]
};
 
var other = {
  'a': [{ 'c': 3 }, { 'e': 5 }]
};
 
_.merge(object, other);
// => positional merge { 'a': [{ 'b': 2, 'c': 3 }, { 'd': 4, 'e': 5 }] }

merge(object, other);
// => concat the arrays { 'a': [{ 'b': 2, 'd': 4, 'c': 3, 'e': 5 }] }
```

Fix #1269

### DIFF
--- a/src/reconciler/proxies.js
+++ b/src/reconciler/proxies.js
@@ -3,8 +3,7 @@ import { resetClassProxies } from '../proxy/createClassProxy';
 import { isCompositeComponent, isReactClass } from '../internal/reactUtils';
 import configuration from '../configuration';
 import { incrementHotGeneration } from '../global/generation';
-
-const merge = require('lodash/merge');
+import { merge } from './utils';
 
 let signatures;
 let proxiesByID;
--- a/src/reconciler/utils.js
+++ b/src/reconciler/utils.js
@@ -74,3 +74,25 @@ export const areSwappable = (a, b) => {
   }
   return false;
 };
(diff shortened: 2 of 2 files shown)


KidkArolis:
Reading the comments further in #1269, I'll see if I can remove merge entirely..

theKashey:
:killitwithfire:

PR #1293, opened 2019-07-07
fix: babel 7.5, fixes #1292

### LINKED ISSUE #1292: Compatibility with babel@7.5.0: Don't break the default exported statement
### Description

RHL makes the  'export default' statement omitted while using @babel/plugin-transform-typescript@7.5.1.

### Expected behavior

Keep the 'export default' statement.

### Environment
```json
{
  "react-ho-loader": "^4.12.3",
  "@babel/core": "^7.5.0",
  "@babel/cli": "^7.5.0",
  "@babel/plugin-transform-typescript": "^7.5.1"
}
```

### Reproducible Demo

Input:
```typescript
export default 'foo';
```

Output:
```javascript
(function () {
  var enterModule = (typeof reactHotLoaderGlobal !== 'undefined' ? reactHotLoaderGlobal : require('react-hot-loader')).enterModule;
  enterModule && enterModule(module);
})();

var __signature__ = typeof reactHotLoaderGlobal !== 'undefined' ? reactHotLoaderGlobal.default.signature : function (a) {
  return a;
};

const _default = 'Foo';
; // 'export default _defualt'' was Omitted <<<<<<<<<<<<<<<

(function () {
  var reactHotLoader = (typeof reactHotLoaderGlobal !== 'undefined' ? reactHotLoaderGlobal : require('react-hot-loader')).default;

  if (!reactHotLoader) {
    return;
  }

  reactHotLoader.register(_default, "default", "/Users/qiqiboy/develop/tigerNewTest/babel-debug/test.tsx");
})();

;

(function () {
  var leaveModule = (typeof reactHotLoaderGlobal !== 'undefined' ? reactHotLoaderGlobal : require('react-hot-loader')).leaveModule;
  leaveModule && leaveModule(module);
})();
```

### The related issue

https://github.com/babel/babel/issues/10162#issuecomment-508985150

> https://github.com/gaearon/react-hot-loader/blob/master/src/babel.dev.js#L118
insertBefore returns an array of ast paths, and that plugin should call path.scope.registerDeclaration(resultOfInsertBefore[0])

Maybe I should open a PR, but I don't know how to do it.

theKashey:
Related:
- https://github.com/babel/babel/issues/10162
- https://github.com/gatsbyjs/gatsby/issues/15423

I actually have to idea what this "fix" is fixing, but...

### DIFF
--- a/src/babel.dev.js
+++ b/src/babel.dev.js
@@ -115,7 +115,9 @@ function plugin(args, options = {}) {
         const expression = t.isExpression(path.node.declaration)
           ? path.node.declaration
           : t.toExpression(path.node.declaration);
-        path.insertBefore(t.variableDeclaration('const', [t.variableDeclarator(id, expression)]));
+        path.scope.registerDeclaration(
+          path.insertBefore(t.variableDeclaration('const', [t.variableDeclarator(id, expression)]))[0],
+        );
         path.node.declaration = id; // eslint-disable-line no-param-reassign
 
         // It won't appear in scope.bindings


pieh:
I have absolutely no context on change, but will just share how this change impacts reproduction I use:

Here's input file: https://github.com/gatsbyjs/gatsby/blob/master/examples/using-typescript/src/layouts/index.tsx

And here's how applying change from this PR makes impact on output: https://www.diffchecker.com/wya2XyXe

qiqiboy:
👍 Fix #1292

nicolo-ribaudo (approved):
(Babel team member here)

This PR is correct.

I'm sorry if v7.5 of `@babel/plugin-transform-typescript` broke compatibility with this plugin, but the only way of implementing many features in Babel is by assuming that the scope information is correct. I think that this missing `registerDeclaration` plugin never caused issues because the only transformation that ever happened to that `const _default` declaration was to transform it to `var`, and that plugin had been designed not to fail in case of missing bindings in the scope.

nicolo-ribaudo:
I'm working on a fix on Babel side not to rely on scope information. I think that this PR should be merged anyway isnce it does the correct thing, but at least we won't break compatibility with older `react-hot-loader` versions.

pieh:
Huge thanks @nicolo-ribaudo and @theKashey for fast turnaround on this ❤️

PR #1304, opened 2019-07-17
fix: script error on IE11 due to lack of Object.entries

steven-qi:
Usage of `Object.entries` (by https://github.com/gaearon/react-hot-loader/pull/1288) cause script error on IE11 due to lack of support.

Use `Object.keys` instead, as `Object.keys` has been used multiple times within the codebase, and has better browser support.

### DIFF
--- a/src/reconciler/utils.js
+++ b/src/reconciler/utils.js
@@ -84,9 +84,8 @@ export function merge(...sources) {
       }
       acc = [...acc, ...source];
     } else if (source instanceof Object) {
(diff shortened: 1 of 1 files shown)


steven-qi:
@KidkArolis please take a look. This is caused by commit https://github.com/gaearon/react-hot-loader/commit/5185464c6ed2099c8970ec8fbf87663c916a209c

theKashey:
👍

PR #1306, opened 2019-07-22
Update Context Provider, fixes #1207

### LINKED ISSUE #1207: React context lost on reload when loading it with useContext hook
After hot reload, useContext no longer returns the context, I assume the whole context is lost, haven't tried it with a normal consumer.

theKashey:
Bug: 4.12+ versions, with `hot-loader/react-dom` enables are not properly handling React.Context if `integratedResolver` is enabled (and it is)

Details: Context values are reset to the defaults.

Reason: "new" context is used to read data from, "old" to provide data. So there is no "new" data.

Fix: update fiber.

Probably better fix: split `resolver` into two parts - to still wrap `lazy`, `context` and other things with proxies, but not wrapping `classes` and `FC`s is it's possible not to wrap.

### DIFF
--- a/examples/styled-components/src/Counter.js
+++ b/examples/styled-components/src/Counter.js
@@ -7,7 +7,8 @@ const ComponentA = () => {
   const [state] = useState('A');
   return (
     <div>
-      {state}-{value}
+      {state}-{value}-
+      <TimerContext.Consumer>{v => v}</TimerContext.Consumer>
     </div>
   );
 };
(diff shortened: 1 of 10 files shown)


theKashey:
The real fix - use `resolveType` to get a proper Context in `useContext`.

theKashey:
Broken since - v4.11.0, for about 20 days.

Reason - RHL wraps `Context` providing the "first registered" version of `Context.Provider`, while `useContext` is using a local variable to read data from.

Solution:
- use `resolveType` in `useContext` to lead to the right, actually used, context variable
- dont wrap Context of react-hot-dom patch is landed.

PR #1308, opened 2019-07-25
update Code Splitting config

aibeb:


theKashey:
It's also worth to mention that no special treatment is required for `React.lazy`

aibeb:
@theKashey 
I used react lazy to implement code splitting, using the new hot api.

The code is as follows. I can also give an example to the examples directory if needed.

`import { hot } from 'react-hot-loader/root';`
`import React, { lazy, Suspense } from 'react';`

`const Home = lazy(() => import("./pages/Home"));`

`const App = () => {
  return (
    <Suspense fallback={<div>Loading...</div>}>
		<Home /></Suspense>)}`

`export default hot(App);`

theKashey:
It's more about - you don't have to change anything inside `"./pages/Home"` in case of `React.lazy`, but you have to if you are using [loadable-componets](https://github.com/smooth-code/loadable-components) for example or update would not happen.

aibeb:
I tried the following configuration and can update automatically.

`import Loadable from 'react-loadable';`
`import Loading from './components/Loading';`

`const SignIn = Loadable({
	loader: () => import('./pages/SignIn'),
	loading: Loading,
});`

`const App = () => { return ( <SignIn />)}`

`export default hot(App);`

PR #1313, opened 2019-07-27
Add an example of loading the patch in webpack

silvenon:


theKashey:
👍, let's make it more clear.

PR #1314, opened 2019-07-29
Clarify docs for @hot-loader/react-dom

silvenon:
By using the name of the package `@hot-loader/react-dom` instead of "React-🔥-Dom" right away we're making it more obvious and accessible what we're talking about; people will be able to instantly search for that package instead of wondering what it is and where it's explained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ export default hot(App);
   }
   ```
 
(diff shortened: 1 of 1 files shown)


theKashey:
Could only strongly agree 😅

PR #1320, opened 2019-08-09
React 16.9.0 patch support

### LINKED ISSUE #1319: React 16.9.0 support webpack plugin
Due to changes in react 16.9, the patch in the webpack plugin no longer works
Also, seems like they baked react-refresh code into react-dom, that is very similar to this code,
So maybe with some tweaks this patch won't be needed at all?

The change:
```js
if (child.tag === Fragment ? element.type === REACT_FRAGMENT_TYPE : child.elementType === element.type)
```
```js
        if (child.tag === Fragment ? element.type === REACT_FRAGMENT_TYPE : child.elementType === element.type || (
        // Keep this check inline so it only runs on the false path:
        isCompatibleFamilyForHotReloading(child, element))) 
```

Related: hot-loader/react-dom#14

Bnaya:
Fixes #1319

I didn't really checked it. just the rest are passing (Or just failing as the few fails on master)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["orta.vscode-jest", "editorconfig.editorconfig", "dbaeumer.vscode-eslint"]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "debug.node.autoAttach": "on"
+}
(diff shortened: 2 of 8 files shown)


Bnaya:
After re-looking at the tests - its not same tests that failing now,
And less tests are running.
Closing this PR for now

Bnaya:
I've re-opened it for the CI checks.
I've made small changes to the test suite, and also excluded react 15
I don't know what's the issue with REACT 15 or with the missing throws.
Probably you will know better

theKashey:
I'll take a look.

PR #1343, opened 2019-09-16
Make typescript-no-babel example work with latest react-hot-loader environment

maratfakhreev:
I've found that typescript-no-babel example doesn't work from scratch and even more fun has babel in dependencies. So I've cleaned the code, updated dependencies and have refactored webpack.config

### DIFF
--- a/examples/typescript-no-babel/.babelrc
+++ b/examples/typescript-no-babel/.babelrc
@@ -1,4 +1,0 @@
-{
-  "presets": ["env", "react"],
-  "plugins": ["transform-class-properties"]
-}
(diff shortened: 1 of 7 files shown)


theKashey:
React Hot Loader still very need `babel` to work out sources. Without it many features, hooks reloading included would not work.

theKashey:
Ok, this was`no-babel` case. Just wondering - should it be deprecated, as long as it's guaranteed to work not properly.

maratfakhreev:
I've tested it and it works correct with current version of rhl and environment, so I'm not sure about it. It works fine with `awesome-typescript-loader` but if you think that in the future it can provides problems maybe yeah, it should be deprecated

theKashey:
I should not stand changing the hook order, but it always was not standing it, so - changed you've provided are the improvement in any case.

PR #1344, opened 2019-09-16
Make typescript example work with latest react-hot-loader environment

maratfakhreev:
As in #1343 have made fixes. Have refactored webpack.config and updated dependencies. Previously It didn't work from scratch and has a lot of warnings.

### DIFF
--- a/examples/typescript/package.json
+++ b/examples/typescript/package.json
@@ -6,26 +6,27 @@
     "start": "webpack-dev-server --hot"
   },
   "devDependencies": {
-    "@babel/core": "^7.0.0",
-    "@babel/plugin-proposal-class-properties": "^7.0.0",
-    "@babel/plugin-proposal-decorators": "^7.0.0",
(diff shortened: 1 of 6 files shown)


theKashey:
Just wondering about your opinion of using `babel` to transpire `TS`. I personally stuck with `ts+babel` due to code indigestible by babel patterns (like not using `isolatedModules`), however many people don't have these problems, and prefer to using a single complier.

What do you think about it - should this example be babel based, or ts+babel based? (or we should add one more example?)

maratfakhreev:
You're asking about loaders? To be honest I didn't find such problem so could you please provide more details? I'm using babel-loader but have had build with ts-loader as well and never mixed them.

Anyway I think the main goal of examples to give simple config that works with rhl without problems. And any other issues go beyond this.

As a result for simplicity I think this example should be babel based only.

theKashey:
Then - merged. Thank you for your contribution.

Issue #1329, opened 2019-08-28
Electron and react-hot-loader boilerplate


nahuelarjonadev:
### Description

Please, I'm desperate. For the last 3 days, I've been trying without luck to implement react-hot-loader within an electron app and don't know what else to try. I'm following many comments, posts, issues trying to replicate what they did but none gives me enough information about the entire setup so I can find what I'm missing. I've already tried electron-webpack but it didn't work for me. It has HMR from webpack, but no react-hot-loader and that makes applications using react-router a pain to work with. Maybe it's not possible, but people seem to have achieved it so I'm not willing to give up.

### Expected behavior

Project setup with electron, webpack, webpack-dev-server, react, react-router, react-hot-loader. I want to be able to change CSS and HTML strings without having to rebuild the entire project with webpack and restart it.

### Actual behavior

Looking for the right styles in a component accessed through react-router links takes forever. Can't keep doing this. I've been using the dev tools for that purpose but it's not even close to the speed achievable by trying things in an IDE.

### Environment

React Hot Loader version: "^4.12.12"
node: 10.16.2
npm: 6.9.0

Operating system: Ubuntu 18.04 LTS
Browser and version: electron "^4.2.9" 

### Reproducible Demo

This is the project I'm trying to set it up for [https://github.com/nahuelarjonadev/kafka-lens/](https://github.com/nahuelarjonadev/kafka-lens/), but I'm ultimately looking for just a super simple working demo so I can mirror the configurations or at least get a hint on how to do it. I couldn't find one yet.

theKashey:
Here is the only issue we had with Electron, and it was working (some time ago) - https://github.com/gaearon/react-hot-loader/issues/794

maratfakhreev:
@nahuelarjonadev @theKashey Couple of weeks ago I've finished my own small app so if you guys are interested I can provide a fresh example on electron and typescript which works with latest react-hot-loader.

theKashey:
That would be great.

mnpenner:
You forgot to include `@hot-loader/react-dom` in `package.json`

https://github.com/gaearon/react-hot-loader/blob/ed8b31080e766ae07c0dd2dc27b639df45cbc0c9/examples/electron-typescript/webpack.config.js#L18

PR #1347, opened 2019-09-19
Introduce electron typescript example which works with react-hol-loader

maratfakhreev:
Regarding to #1329 I've added example of electron app which uses typescript as well. It's really hard to find boilerplate with this too technologies in one place so I think it will be useful for all.
- I've also provided two tasks for development and compiling and enhance webpack config for them.
- As app I've used the counter from typescript example.


theKashey:
I reckon many people would say a `BIG THANK YOU` to you

muscaiu:
Moving `react-hot-loader` and `@hot-loader/react-dom` to devDependencies fixed it for me. 
Using Electron with Electron-React-Redux boilerplate.

PR #1354, opened 2019-10-06
Getting Started guide now in line with up-to-date README

### LINKED ISSUE #891: Update or remove "Get Started" in https://gaearon.github.io/react-hot-loader/getstarted/
The get started section in https://gaearon.github.io/react-hot-loader/getstarted/ seems to be outdated in comparison to the readme. It would be better to remove it or link it to the readme.

Antrikshy:
This resolves #891.

I started off by gutting most of the existing content on the Getting Started page, then copy pasted a barebones excerpt of what seemed right from the main README, followed by a link to the README.

I'm not sure how important hooks support is, as I've never personally set that up in a project myself. Let me know if I should drop the `@hot-loader/react-dom` section too and keep it even leaner.

Also tweaked the CSS a teeny bit to get rid of the "Code Code" on each code snippet on the site. I couldn't figure out why there were two of them, so I got rid of both to reduce clutter.

### DIFF
--- a/getstarted.md
+++ b/getstarted.md
@@ -77,4 +77,6 @@ module.exports = {
 
 ## More documentation
 
-For more detailed documentation, see [README on GitHub](https://github.com/gaearon/react-hot-loader/blob/master/README.md).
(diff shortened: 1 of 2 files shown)


Antrikshy:
Just found the troubleshooting doc, so I added a link to that at the bottom, along with a "contributions welcome" message.

PR #1358, opened 2019-10-11
Added @hot-loader/react-dom which is missing

theajr:
@gaearon 

In `examples/typescript`, `@hot-loader/react-dom` was missing in `package.json` which is required as we are using it as alias for react-dom.
Fixed some auto fixable vulnerabilities by running `npm audit fix`


theKashey:
`package.lock` is not supposed to be here 😅

theajr:
@theKashey Removed package lock file!

theKashey:
Good job!
There is another TypeScript related issue you might like - https://github.com/gaearon/react-hot-loader/issues/1359 (but you don't have to)

theajr:
Sure, I will check it out!

PR #1367, opened 2019-10-25
Update troubleshooting doc with new tip

CommanderHamster:
Setting `NODE_ENV` to `production` seems obvious that it would turn hot-reloading off, however setting `NODE_ENV` to `test` seems much less obvious and worth a call out.

This is an issue that I ran into with my project setup that was hard to track down. First I wasn't sure why the components didn't know how to hot reload since I followed the examples and had wrapped them properly with `export default hot(Component)`. There wasn't any messaging to tell me that `react-hot-loader` was running in production mode, so it was hard to figure out. After looking through all the documentation (including this file) and exhausting everything I found there I looked into the source code and found that when `NODE_ENV === 'test'`, react-hot-loader runs in production mode. I would have had a much easier time figuring it out had there been a tip here, so I thought I should add it for anyone that runs into this in the future.


theKashey:
👍 even I forgot about the `test` case.

PR #1385, opened 2019-10-30
Create README.md

TotooriaHyperion:
Warn people that hook reloading is relying on babel, in case someone use google search to find this example.


theKashey:
👍

PR #1387, opened 2019-10-30
fix(track-trail-updates): added ability to turn off trailing updates …

### LINKED ISSUE #1386: Suppress trackTailUpdates if 'true' passed
The docs say `The warning is informational - it is a notice that this tail update logic is triggered, and does not indicate a problem in the configuration or usage of react-hot-loader.`. I have a team of developers working with me, I'm managing the SSR/HMR setup. I don't think every developer should need to know about the inner works of a HMR. Ideally, I can set setConfig({ trackTailUpdates:true }). to suppress the out-of-bounds warning since it's just informational and not an issue with the configuration. I'd like to keep our console logs to a minimal.

Discussion also happening here: #1368

Opening a new issue to track when I create a PR

brandon-pereira:
Closes #1386

### DIFF
--- a/src/hot.dev.js
+++ b/src/hot.dev.js
@@ -66,9 +66,12 @@ const makeHotExport = (sourceModule, moduleId) => {
                 // we know that some components were updated, but not tracking which ones
                 // even if their updates might be incorporated automatically (like lazy)
                 // we dont know which one should be tracked, and which updates are important
-                console.warn(
-                  'React-Hot-Loader: some components were updated out-of-bound. Updating your app to reconcile the changes.',
-                );
+                if (typeof configuration.trackTailUpdates === 'boolean' && configuration.trackTailUpdates) {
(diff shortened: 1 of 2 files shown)


src/hot.dev.js
```diff
@@ -66,9 +66,12 @@ const makeHotExport = (sourceModule, moduleId) => {
                 // we know that some components were updated, but not tracking which ones
                 // even if their updates might be incorporated automatically (like lazy)
                 // we dont know which one should be tracked, and which updates are important
-                console.warn(
-                  'React-Hot-Loader: some components were updated out-of-bound. Updating your app to reconcile the changes.',
-                );
+                if (typeof configuration.trackTailUpdates === 'boolean' && configuration.trackTailUpdates) {
```
theKashey: It's `true` by default, and is a condition to run this branch.

theKashey:
Let's continue in #1389

PR #1404, opened 2019-12-25
Update redux example and remove some file

liuliangsir:
- As the above title goes


theKashey:
Thanks for the cleaning up!

Issue #1001, opened 2018-06-04
🔥Not all methods could be updated


theKashey:
Each component consist of 2 different pieces:
1. Class methods. They exists on prototype and could be updated just by replacing the old prototype by a new one.
2. Class properties. Their initiation occurs in class constructor.

To update second ones RHL creates 2 components - old and new, then compares all descriptors.
Comparison ignores "objects", thus keeps state  untouched, and updates functions and simple variables.

But that does mean "updates functions"?
It literally cast function _toString_ and then eval that code in the context of Class.

# The problem
Casting function toString defines _what is it_, not _how it was created_.
Before, RHL did update, only if "text" got changes, but it was changed https://github.com/gaearon/react-hot-loader/issues/949 and now they always got updated.

This causes https://github.com/gaearon/react-hot-loader/issues/995, https://github.com/gaearon/react-hot-loader/issues/978, https://github.com/gaearon/react-hot-loader/issues/969, https://github.com/gaearon/react-hot-loader/issues/984

RHL have rights only to update "own" functions, but could not detect them.

# Solution

1. Revert #949 
2. Update "arrow" functions only if text got changed. Or if "this" keysword is found inside (probably should fix #949 back)
3. Update `bound` class methods. Ie those ones, which names exists on prototype.

theKashey:
Look like it is impossible to properly "hot" update class, and even heavy babel magic will not work.
We should not try to find a way to fix the problem, which does not exists.

The right way:
1. Create a new class
2. Copy state and all variables you have to keep from old component to  the new one
3. Switch to the new component.

__Not the change current component, but adopt a new one__

This require a one line change in React-Fiber - https://github.com/facebook/react/blob/dd5fad29616f706f484938663e93aaadd2a5e594/packages/react-reconciler/src/ReactChildFiber.js#L375
The similar changes as (not yet) was made to Preact  - https://github.com/developit/preact/pull/1120

theKashey:
fixed in v4.3.0

xinggangling:
nice

theKashey:
To be more concrete - this issue is __NOT CLOSED__.

This issue addresses the problem when class methods __should not be updated__, or this action would break your code - as a result, some methods __will not be updated__, while and __could__, and __should__.
Today this leads to partial incompatibility with React-Redux v6.

PR #1415, opened 2020-01-29
Fix link to issue 1001 in readme

maxgabut:
Use the full URL to link to issue #1001 in the readme.
I think the short form that was used (`#1001`) only works in comments and PR messages.


theKashey:
👍

PR #1421, opened 2020-02-17
Fix styles for code blocks because of nested .highlight elements

silvenon:
For some reason the highlighter output:

```html
<div class="highlight">
  <pre class="highlight">
    <!-- ... -->
  </pre>
</div>
```

which causes all styles with the selector `.highlight` to be applied on both. Most of them don't have a visible impact, except the one which adds a pseudo-element:

Screen Shot 2020-02-17 at 19 01 38

This is now fixed by specifying the element.

### DIFF
--- a/public/css/playground.css
+++ b/public/css/playground.css
@@ -352,7 +352,7 @@ small code,li code,p code {
     text-align: right;
 }
 
-.highlight:after {
+div.highlight:after {
     position: absolute;
     top: 0;
     right: 0;


theKashey:
You've changed how it looks, not how it reads. gh-pages were not properly updated for years.

silvenon:
If you mean that I didn't fix the underlying bug of two `.highlight` elements, that's true, but it's a bug somewhere in dependencies in `github-pages` gem.

theKashey:
I mean that you might have fixed how it looks, now what is written there. The content itself contain a lot of  _"errors"_.

silvenon:
Yeah, at the time I didn't realize that this change is pretty meaningless in comparison to everything that needs to be updated there, but the purpose of this change was only to patch how it looks.

PR #1428, opened 2020-03-01
fix: patch support for react-dom 16.13

theKashey:


theKashey:
Testing with 16.10/12 - shall not break older versions...

PR #1432, opened 2020-03-18
Remove "deprecation" in readme to avoid confusion

Hoishin:
Putting big "Deprecation" title in README is confusing when the replacing library is still [experimental](https://github.com/facebook/react/tree/master/packages/react-refresh#react-refresh).


theKashey:
Yeah. I am still waiting to be deprecated, and every day it's not today :(

PR #1433, opened 2020-03-22
add peerDependenciesMeta in package.json

bertho-zero:


theKashey:
As per https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md
👍 , cool stuff!

Issue #1449, opened 2020-05-08
react-hot-loader may break the app in Webpack 5.0.0-beta.14 and up


wojtekmaj:
Note: This has also been [reported in Webpack repo](https://github.com/webpack/webpack/issues/10854).

### Description

I'm getting an error "ReferenceError: exports is not defined" when running the application that uses with `react-hot-loader` and imports `d3-array` built with Wepack 5.0.0-beta.14 and up.

### Expected behavior

"Uncaught ReferenceError: exports is not defined" not to be thrown when executing Webpack-bundled code.

### Actual behavior

Here's the exact error:

```
Uncaught ReferenceError: exports is not defined
    at register (cross.js?0cbf:23)
    at eval (cross.js?0cbf:23)
    at Object.../node_modules/d3-array/src/cross.js (charts.js:376)
    at __webpack_require__ (src.js:6070)
    at fn (src.952fd4b1.js:6330)
    at eval (index.js:18)
    at Object.../node_modules/d3-array/src/index.js (charts.js:496)
    at __webpack_require__ (src.js:6070)
    at fn (src.js:6330)
    at eval (band.js:5)
```

Any of the following seem to unbreak the app:

* removing `d3-array` import (duh)
* `export default hot(Root)` -> `export default Root`
* applying `react-hot-loader/webpack` loader to `/node_modules\/react-dom/` only and not `/node_modules/`

Tested other versions of Webpack, changing absolutely nothing but the Webpack version, and here's my findings:

Version | Crashes?
-- | --
4.43.0 | ✅ No
5.0.0-beta.12 | ✅ No
5.0.0-beta.13 | ✅ No
5.0.0-beta.14 | ❌ Yes
5.0.0-beta.16 | ❌ Yes

### Environment

React Hot Loader version: 4.12.21

Run these commands in the project folder and fill in their results:

1.  `node -v`: 12.16.2
2.  `npm -v`: 6.14.4

Then, specify:

1.  Operating system: macOS Catalina
2.  Browser and version: Chrome 81.0.4044.129

### Reproducible Demo

**[Reproducible repo](https://github.com/wojtekmaj/webpack-crash)**

theKashey:
> applying react-hot-loader/webpack loader to /node_modules\/react-dom/ only

Is a good idea. There is no need to apply any "hot" stuff on the "static" node_modules

Additionally, you can pass `noRegister` option to webpack loader and it will skip the moment which is failing.

[...]

PR #1452, opened 2020-05-11
Discourage using webpack loader for all the files

wojtekmaj:
* Replaces the phrase "It is safe to enable this loader for all the files" with "It is _possible_ to enable...", because it's neither safe (as proven in #1449) nor usually necessary.
* Switches first example of Webpack config (suboptimal, applying loader to all the files) with the second one (optimal, applying loader to react-dom only, if you're using babel plugin)

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,16 +382,17 @@ class MyComponent extends React.Component {
 
 But `webpack-loader` could help with TypeScript or _spreading_ "cold API" [to all node_modules](https://github.com/gaearon/react-hot-loader#disabling-a-type-change-for-all-node_modules).
 
-> It is safe to enable this loader for all the files. But place it after babel-loader, if babel-loader is present.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -401,13 +401,13 @@ module.exports = {
 
 Webpack plugin will also land a "hot" patch to react-dom, making React-Hot-Loader more compliant to [the principles](https://github.com/gaearon/react-hot-loader/issues/1118).
 
-If you are using `babel-plugin` you might not need to apply `webpack-loader` to all the files, scoping it to `react-dom`
+If you are not using `babel` plugin you might need to apply `webpack-loader` to all the files.
 
 ```js
 // would only land a "hot-patch" to react-dom
```
theKashey, line 407: this comment should be moved to the code example above.

wojtekmaj: Yep my bad! Thanks, fixed.

theKashey:
👍

Issue #1453, opened 2020-05-18
Deprecate React-Hot-Loader


theKashey:
Deprecate to `React-refresh`

Requires 3 parts -> fast-refresh, babel, webpack.

- (react, babel) https://github.com/facebook/react/tree/master/packages/react-refresh
- (wp4) https://github.com/pmmmwh/react-refresh-webpack-plugin
- (wp5) https://github.com/WebHotelier/webpack-fast-refresh
- (wp4) https://github.com/zeit/next.js/tree/master/packages/react-refresh-utils

- no solution for Electron
- no solution for Parcel

theKashey:
https://github.com/gaearon/react-hot-loader/pull/1513

PR #1454, opened 2020-05-19
@next refers old version (v4.5.3)

mtsmfm:
https://www.npmjs.com/package/react-hot-loader

image

### DIFF
--- a/getstarted.md
+++ b/getstarted.md
@@ -160,7 +160,7 @@ To preserve *internal component state*, you now need to add `react-hot-loader` t
 1.  Install the package:
 
     ```sh
-    $ npm install --save-dev react-hot-loader@next
+    $ npm install --save-dev react-hot-loader
     ```
 1.  Add the package to your config:
 


theKashey:
👍

mtsmfm:
@theKashey next is currently not next and beta is not beta I think.
Do you have any plans to remove these tags?

theKashey:
There is a plan to remove RHL :) Not joking - this package is very close to be deprecated.

mtsmfm:
I know this issue https://github.com/gaearon/react-hot-loader/issues/1453 though,
I think react-refresh is still experimental and we'll need some months to deprecate.

theKashey:
It's already used in `nextjs` and `react-native`, and even RHL uses it from some point of view.

mtsmfm:
I see.

Anyway, this package is not deprecated yet, removing old tags can avoid confusing from now I think.

https://docs.npmjs.com/cli/dist-tag

PR #1708, opened 2021-06-01
Note create-react-app support for fast refresh in readme

loren138:
Adds a link and details for using fast refresh now native in create-react-app since version 4.


theKashey:
Can you move it below the "bundlers" section and add nextjs as well.

loren138:
@theKashey I've added next.js and moved them down the list.  Let me know if there is anything else.

theKashey:
👍

Issue #1442, opened 2020-04-23
loader-utils vulnerability upgrade


niryarden:
Hello :)

loader-utils dependency released a new version (2.0.0) about a month ago, in which they fixed a vulnerability following a snyk report. I would appreciate if you could update loader-utils to the latest version.

webpack/loader-utils#165
https://app.snyk.io/vuln/SNYK-JS-MINIMIST-559764

Thank you very much in advance

Shivam60:
Hey Any update on this

Shivam60:
I would like to help also, if its a good for first bug

mateBe95:
bump

psierks:
There's also [CVE-2022-37601](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2022-37601)

loader-utils 1.0.0 to 2.0.2 are affected, patched as of 2.0.3

loader-utils fix PR https://github.com/webpack/loader-utils/pull/217 & release https://github.com/webpack/loader-utils/releases/tag/v2.0.3

dodo0822:
Following. Would appreciate if we could remove the dependency on loader-utils by dropping support for webpack 4.

NidhiLearning:
I have updated the react-hot-loader dependency with version with version 4.13.0 but still I am getting the older loader-utils version in the package.json. Any suggestions? Is there anything which I am missing?

theKashey:
@NidhiLearning - please use 4.13.1

PR #1849, opened 2022-10-26
update loader-utils version

psierks:
- Bumping loader-utils version

Addresses https://github.com/gaearon/react-hot-loader/issues/1442

Seems `getOptions` is the only used export from loader-utils
https://github.com/gaearon/react-hot-loader/blob/f58a9319266cac630119727e8d681f3a5111348b/src/webpack/index.js#L35
And while it was removed in [v3](https://github.com/webpack/loader-utils/blob/v3.0.0/lib/index.js), it's still available in [v2.0.3](https://github.com/webpack/loader-utils/blob/v2.0.3/lib/index.js#L14) which contains the [fix](https://github.com/webpack/loader-utils/pull/217) for [CVE-2022-37601](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2022-37601)

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "fast-levenshtein": "^2.0.6",
     "global": "^4.3.0",
     "hoist-non-react-statics": "^3.3.0",
-    "loader-utils": "^1.1.0",
+    "loader-utils": "^2.0.3",
     "prop-types": "^15.6.1",
     "react-lifecycles-compat": "^3.0.4",
     "shallowequal": "^1.1.0",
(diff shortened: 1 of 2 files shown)


psierks:
Just seeing https://github.com/gaearon/react-hot-loader/pull/1848/files
Guess it's time to switch away from react-hot-loader

theKashey:
I would really like to deprecate RHL, and do encougare switching to modern solutions -  as long as there is a million monthly downloads - I cannot do that.

In other words 👍 thank you for your contribution

